PR #1, opened 2018-08-06
RIA-305 Hello World

fantasticjamieburns:


Jenkinsfile_CNP
```diff
@@ -0,0 +1,16 @@
...
+        pipelineTriggers([[$class: 'GitHubPushTrigger']])
+])
+
+@Library("Infrastructure")
+
+def type = "java"
+def product = "ia"
+def component = "case-api"
+
+withPipeline(type, product, component) {
```
jackmaloney: Take it you don't need to point at Azure Vault at this stage?

fantasticjamieburns: Yeah we are not setup in CNP so rather wait until somebody says we have a vault.

PR #7, opened 2018-09-12
Use service plan

danielepolencic:
In a few words: we currently use a service plan for application. After the change, we will have a single plan per project.

While we currently have a 1:1 mapping between applications and compute resources, with a shared plan we can fit more apps into a single compute resource.
The change is rolled out in two stages.

1. Explicitly enable the app service plan in Terraform
1. Switch the service plan to use a shared service plan.

**PLEASE NOTE, THIS PR IS FOR STEP 1 ONLY**

More info on the change: https://tools.hmcts.net/confluence/display/CNP/Migrate+to+Shared+App+Service+Plan (please note that the instructions are for the full migration).

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -17,7 +17,7 @@ data "azurerm_key_vault_secret" "ia_case_api_url" {
 }
 
 locals {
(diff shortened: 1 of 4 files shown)


fantasticjamieburns:
Closing PR to get a [PREVIEW] build to test out the terraform

PR #13, opened 2018-10-29
RIA-473 Add serenity

Rob-Doherty:
### WHAT?

This change adds Serenity reporting to the existing RestAssured tests.

### WHY?

Serenity creates a really clear test report as it's output, which should help pin-point failures (and hopefully their reasons).

### HOW?

Serenity was included as dependency for the project and incorporated in the functionalTest. The Jenkinsfiles were updated to publish the reports as part of the builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .gradle
 /build/
+/target/
 !gradle/wrapper/gradle-wrapper.jar
 
 ### STS ###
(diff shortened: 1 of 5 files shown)


Rob-Doherty:
Waiting for info on whether merging the serenity reporting is ok (see JIRA RIA-473)

Rob-Doherty:
Have had confirmation (see JIRA RIA-473) that this is safe to merge.

PR #14, opened 2018-11-02
[PREVIEW] RIA-473 make Jenkins archive serenity reports correctly

Rob-Doherty:
### WHAT?

Serenity reports were not being correctly archived by Jenkins, and so this change fixes that. 

### WHY?

Otherwise any failures that have been correctly reported by Jenkins would not be available for viewing without this change


Rob-Doherty:
Test reports now being exposed correctly by Jenkins job

PR #16, opened 2018-11-13
RIA-730 Home Office decision number

fantasticjamieburns:


src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/util/StringResourceLoader.java
```diff
@@ -0,0 +1,45 @@
...
+        // noop
+    }
+
+    public static Map<String, String> load(String locationPattern) throws IOException {
+
+        Resource[] resources =
+            new PathMatchingResourcePatternResolver()
+                .getResources(locationPattern);
+
+        return
```
t80: Could just be my subjectivity but this stream seems hard to read

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/config/AuthCheckerConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Bean
+    public Function<HttpServletRequest, Collection<String>> authorizedRolesExtractor() {
+        return (any) -> ImmutableSet.copyOf(authorisedRoles);
+    }
+
+    @Bean
+    public Function<HttpServletRequest, Optional<String>> userIdExtractor() {
+        return (any) -> Optional.empty();
```
t80: Intentionally returning Optional.empty();

src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/util/StringResourceLoader.java
fantasticjamieburns: Fair enough -- I will take the anonymous function and extract it into a method which should make it all read better 👍

src/smokeTest/java/uk/gov/hmcts/reform/iacaseapi/SmokeTest.java
```diff
@@ -16,20 +17,20 @@
...
 
         RestAssured.baseURI = targetInstance;
         RestAssured.useRelaxedHTTPSValidation();
 
-        String response = RestAssured
-            .given()
+        String response = SerenityRest
             .when()
-            .get("/")
+            .get("/health")
```
satyachundur: we are already verify health as a part of the deployment.. it would be worth to check any simple get if we have any.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/field/AddressUk.java
```diff
@@ -0,0 +1,87 @@
...
+    private Optional<String> county = Optional.empty();
+
+    @JsonProperty("PostCode")
+    private Optional<String> postCode = Optional.empty();
+
+    @JsonProperty("Country")
+    private Optional<String> country = Optional.empty();
+
+    private AddressUk() {
+        // noop -- for deserializer
```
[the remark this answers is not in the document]
fantasticjamieburns: Default (no args) constructors are used by deserializers (jackson in this case) when it builds the object from JSON. It is marked as private so the default constructor cannot be used by normal application code, only deserializer libraries that build an object using reflection internally.

src/smokeTest/java/uk/gov/hmcts/reform/iacaseapi/SmokeTest.java
fantasticjamieburns: Right now, we do not have a GET resource (except health) that can be accessed without security tokens -- all other application endpoints are secure.

fantasticjamieburns: What we could do is remove security from this controller's endpoint:
https://github.com/hmcts/ia-case-api/blob/master/src/main/java/uk/gov/hmcts/reform/iacaseapi/controllers/RootController.java#L26
To allow it to be useful to the smoke test, but previously Rob said testing the welcome message was not a useful smoke test.

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/serialization/AsylumCaseCallbackDeserializer.java
```diff
@@ -34,7 +30,6 @@ public AsylumCaseCallbackDeserializer(
             );
 
         } catch (IOException e) {
-            LOG.warn("Could not deserialize callback:\n{}\n{}", e.getMessage(), source);
```
raghera: Any reason why the logs in exceptions were removed?  I ask because sonar sometimes complains about not logging errors.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseBuilder.java
```diff
@@ -0,0 +1,156 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.field.AddressUk;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.field.IdValue;
+
+@Component
+public class AsylumCaseBuilder {
```
raghera: I often write "Builder" pattern implementations so client code can chain the methods which for me at least makes it slightly easier to use and more fluent.  Matter of opinion I guess!

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseBuilderTest.java
```diff
@@ -0,0 +1,69 @@
...
+    private final Optional<String> hasOtherAppeals = Optional.of("L");
+    private final Optional<List<IdValue<String>>> otherAppeals = Optional.of(Arrays.asList(new IdValue<>("1", "M")));
+    private final Optional<String> legalRepReferenceNumber = Optional.of("N");
+
+    private AsylumCaseBuilder asylumCaseBuilder = new AsylumCaseBuilder();
+
+    @Test
+    public void should_build_asylum_case() {
+
+        asylumCaseBuilder.setHomeOfficeReferenceNumber(homeOfficeReferenceNumber);
```
raghera: Testing getters/setters can sometimes be a pain.  In the past I have used something like SmartUnit to make it less painful through reflection.  https://github.com/rlogiacco/SmartUnit/wiki/Testing-POJOs-properties.
Think there is another lib called OpenPojo or similar which I have not used.

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/serialization/AsylumCaseCallbackDeserializer.java
fantasticjamieburns: Yeah Sonar said no to logging and throwing in the same block

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseBuilderTest.java
fantasticjamieburns: We are expected to use the same testing suite across all reform projects -- so that is basically JUnit and Mockito. If I had my way we would be writing beautiful Spock tests :) So I don't think we can introduce this tool.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseBuilder.java
fantasticjamieburns: Yeah let's do that, but as it is not a functional change can we leave it to another story -- keen to get this one over the line so I can rebase future work from this PR?

raghera: Yes that's fine, I think getting this done is more important.

PR #18, opened 2018-11-16
RIA-516 New matters changed to a text field

fantasticjamieburns:


fantasticjamieburns:
@Rob-Doherty @t80 @raghera Can I get a reapproval now master is updated?

fantasticjamieburns:
Because these have been approved previously, and because all the tests still pass, I am pushing to master...

PR #26, opened 2018-11-26
RIA-694 Send direction (replay build)

fantasticjamieburns:


fantasticjamieburns:
Migrated from https://github.com/hmcts/ia-case-api/pull/19

PR #30, opened 2018-11-28
RIA-230 & RIA-236 Grounds of appeal

fantasticjamieburns:


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/CheckValues.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.List;
+
+public class CheckValues<T> {
```
raghera: Just for my understanding, what is this class's purpose?  To hold a non null List?

fantasticjamieburns: Ccd gives us the data like this for the Grounds fields:
https://github.com/hmcts/ia-case-api/pull/30/files#diff-3589128de80e18359dbe7becc98422aeR11
So it is a representation of that type of field -- a "Collection of a Complex type consisting of checkbox values".

fantasticjamieburns: Not so much about a non null list -- but a list that is within an object having a single key.

PR #31, opened 2018-11-30
RIA-480 & RIA-600 Build case (appeal argument)

fantasticjamieburns:


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/DocumentWithDescription.java
```diff
@@ -17,16 +18,19 @@ public DocumentWithDescription(
         Document document,
         String description
     ) {
-        this.document = document;
-        this.description = description;
+        requireNonNull(document);
+        requireNonNull(description);
+
+        this.document = Optional.ofNullable(document);
+        this.description = Optional.ofNullable(description);
```
t80: Are the `.ofNullable` required when we've run them through `requireNonNull()`?

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/DocumentWithDescription.java
```diff
@@ -17,16 +18,19 @@ public DocumentWithDescription(
...
     }
 
-    public Document getDocument() {
+    public Optional<Document> getDocument() {
         requireNonNull(document);
```
t80: is `requireNonNull` needed? Its set to `Optional` instance at the field level so can it ever be null?

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/DocumentWithDescription.java
fantasticjamieburns: Shameful display! Not they are not required.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/DocumentWithDescription.java
fantasticjamieburns: Deserializers can bypass constructor so yes they can still be `null`

PR #32, opened 2018-11-30
RIA-517 Submit case argument

fantasticjamieburns:


src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/postsubmit/CaseSubmittedConfirmationTest.java
```diff
@@ -0,0 +1,89 @@
+package uk.gov.hmcts.reform.iacaseapi.domain.handlers.postsubmit;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
raghera: We are using assertJ :)

PR #39, opened 2018-12-13
RIA-610 Send build case notification

fantasticjamieburns:


src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/CcdScenarioRunnerTest.java
```diff
@@ -45,16 +43,44 @@ public void setUp() {
...
+        Collection<String> scenarioSources =
+            StringResourceLoader
+                .load("/scenarios/" + scenarioPattern)
+                .values();
 
-            Headers authorizationHeaders = getAuthorizationHeaders(scenario);
-            String requestBody = buildCallbackRequestBody(scenario, templatesByFilename);
+        System.out.println((char) 27 + "[36m" + "-------------------------------------------------------------------");
+        System.out.println((char) 27 + "[33m" + "RUNNING " + scenarioSources.size() + " SCENARIOS");
+        System.out.println((char) 27 + "[36m" + "-------------------------------------------------------------------");
```
t80: I'm learning good stuff here

src/integrationTest/java/uk/gov/hmcts/reform/iacaseapi/docs/SwaggerPublisher.java
```diff
@@ -45,6 +46,12 @@ public void shouldGenerateDocs() throws Exception {
 
         LOG.info("Generating Swagger Docs");
 
+        File linuxTmpDir = new File("/tmp");
+        boolean linuxTmpDirExists = linuxTmpDir.exists();
+        if (!linuxTmpDirExists) {
```
t80: I think it reads just as well inlined and with one less line of code, and I got to write a review comment, yay!

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/callback/PreSubmitCallbackResponse.java
```diff
@@ -10,11 +10,15 @@
 
 public class PreSubmitCallbackResponse<T extends CaseData> {
 
-    private final T data;
-    private final Set<String> errors = new LinkedHashSet<>();
+    private T data;
+    private Set<String> errors = new LinkedHashSet<>();
```
t80: Just curious about the use a LinkedHashSet here as apposed to standard hashset? And why it went from from final to non-final?

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AppealReferenceNumberHandler.java
```diff
@@ -32,31 +32,31 @@ public boolean canHandle(PreSubmitCallbackStage callbackStage, Callback<AsylumCa
...
+            new PreSubmitCallbackResponse<>(asylumCase);
 
         if (!asylumCase.getAppealReferenceNumber().isPresent()) {
 
             Optional<String> appealReferenceNumber =
                 appealReferenceNumberGenerator.getNextAppealReferenceNumberFor(
-                        asylumCase.getAppealType()
-                            .orElseThrow(() -> new AsylumCaseRetrievalException("Unrecognised asylum case type")));
+                    asylumCase.getAppealType()
+                        .orElseThrow(() -> new AppealReferenceNumberInitializerException("Unrecognised asylum case type")));
```
t80: I'll have to take another look at my formatting, I did import the settings you sent me but obviously something still a miss

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/clients/AsylumCaseNotificationApiSender.java
```diff
@@ -0,0 +1,73 @@
...
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.AsylumCase;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.Callback;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.PreSubmitCallbackResponse;
+import uk.gov.hmcts.reform.iacaseapi.domain.service.NotificationSender;
+import uk.gov.hmcts.reform.iacaseapi.infrastructure.security.UserCredentialsProvider;
+
+@Service
+public class AsylumCaseNotificationApiSender implements NotificationSender<AsylumCase> {
+
+    private static final String SERVICE_AUTHORIZATION = "ServiceAuthorization";
```
t80: Think we've go this is quite a few places now, I'll add to the tech debt to centralize somewhere appropriate

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/JwtAccessTokenDecoder.java
```diff
@@ -17,33 +16,38 @@
...
+                accessToken.replaceFirst("^Bearer\\s+", "")
             );
 
             String accessTokenClaims = new String(
-                    Base64Utils.decodeFromString(jwt.getPayload())
+                Base64Utils.decodeFromString(jwt.getPayload())
             );
 
-            return mapper.readValue(
+            try {
```
t80: Ok I could well be wrong, but I reasoned that technically this second try block-catch wasn't required. the mapper throws when it can't parse the json into a Map<String, String> but we already know its good because JWT.decode (I think) would have thrown already for the same reason.  Also Sonar is suggesting in my IDE that nested try-block is bad

src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/CcdScenarioRunnerTest.java
fantasticjamieburns: I can't tell if you are being serious!
I just wanted the tests to "pop" :)

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/callback/PreSubmitCallbackResponse.java
fantasticjamieburns: LinkedHashSet preserves insert order on iteration, whereas standard HashSet does not -- so the errors should be displayed in the order they were added.

As for the final -- i cannot remember, but what did change about this class is we have started to deserialize it, not just serialize it. I was of the view that final fields and deserialization are not things that work well together:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.5.3

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/clients/AsylumCaseNotificationApiSender.java
fantasticjamieburns: 👍

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/JwtAccessTokenDecoder.java
fantasticjamieburns: Fair points. I will rework it to not be nested. We definitely have to catch `IOException`, else the class will not compile.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/callback/PreSubmitCallbackResponse.java
t80: Yeah my mistake, I was thinking that ordering would be lost after making the copy, just read the docs.

infrastructure/prod.tfvars
```diff
@@ -1,2 +0,0 @@
-capacity = "2"
```
Rob-Doherty: did you definitely want to have deleted

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandlerTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Before
+    public void setUp() {
+
+        sendNotificationHandler =
+            new SendNotificationHandler(notificationSender);
+    }
+
+    @Test
+    public void should_send_notification_and_update_the_case() {
```
Rob-Doherty: Should we be able to send notifications even if the same notification has been sent already? And if so, would we want to cover that in a test?

infrastructure/prod.tfvars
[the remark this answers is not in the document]
fantasticjamieburns: No, I will fix that!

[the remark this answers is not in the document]
fantasticjamieburns: Actually this was correct -- I thought it was the file that was deleted
We only want 1 instance of the API running, as part of the strategy to prevent duplicate emails being sent. Otherwise CCD could send 2 callbacks, one to each instance and we would not be able to prevent 2 emails being sent.

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandlerTest.java
fantasticjamieburns: It is a good thought.

Currently we prevent duplicate emails within a short time (60 seconds) within notifications API, but not over the entire lifecycle of the case.

There maybe cases where the same email may get sent in the future, I am thinking when a case is put back into a previous state for correction. When the case progresses again, another email will likely be needed for the same action taken the second time.

infrastructure/prod.tfvars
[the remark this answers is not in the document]
fantasticjamieburns: And the default capacity is `1`

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandlerTest.java
Rob-Doherty: yeah that would make sense to me, or if there had been some issue and certain notifications needed to be resent. So would it be worth adding that test here, or is that covered by the `ia-case-notifications-api` code?

Rob-Doherty: I think I may have found the answer to my own question here: https://github.com/hmcts/ia-case-notifications-api/blob/8653e8da0b41be6ac673244c3eaabc3b053080a0/src/test/java/uk/gov/hmcts/reform/iacasenotificationsapi/infrastructure/clients/GovNotifyNotificationSenderTest.java#L70

fantasticjamieburns:
Closed and moved to:
https://github.com/hmcts/ia-case-api/pull/55

PR #56, opened 2019-01-14
updated test data: legalRepresentativeName changed to be in line with AAT

Rob-Doherty:
legalRepresentativeName updated to "A Legal Rep" as per AAT test user data


Rob-Doherty:
closing because a cleaner version exists on https://github.com/hmcts/ia-case-api/pull/58

PR #64, opened 2019-01-21
Removing system user until post launch date

fantasticjamieburns:


fantasticjamieburns:
TIDAM working now with system user -- this is not needed (yay)

PR #77, opened 2019-02-18
Ria 369 rescue branch

t80:


src/functionalTest/resources/templates/minimal-appeal-started.json
```diff
@@ -22,5 +22,6 @@
   },
   "hasNewMatters": "No",
   "hasOtherAppeals": "No",
-  "appealReferenceNumber": "DRAFT"
+  "appealReferenceNumber": "DRAFT",
+  "submissionOutOfTime": "No"
```
fantasticjamieburns: Not sure this should exist in `appealStarted` state?

src/functionalTest/resources/templates/empty-case.json
```diff
@@ -1 +1,3 @@
-{}
\ No newline at end of file
+{
+  "submissionOutOfTime": "No"
```
fantasticjamieburns: Not sure this should exist in en empty case

fantasticjamieburns:
Now on master...

PR #80, opened 2019-02-26
[PREVIEW] RIA-1027 Appeal reference numbers generated from DB, not CCD API

fantasticjamieburns:

https://tools.hmcts.net/jira/browse/RIA-1027

### Change description ###

This change replaces the tactical Appeal Reference Number generator.

The tactical solution called into CCD API the first time a reference number was generated. and was not scalable as the number of cases grows, it also meant we could only run a single instance of the Case API so contributed to a less reliable/performant service.

The new solution uses a postgresql database to generate and store the Appeal Reference Numbers and is independent of the CCD API.

```
[ ] Yes
[X] No
```


build.gradle
```diff
@@ -83,6 +84,17 @@ sourceSets {
...
 
+idea {
+    module {
+        testSourceDirs += project.sourceSets.integrationTest.java.srcDirs
+        testSourceDirs += project.sourceSets.integrationTest.resources.srcDirs
+        testSourceDirs += project.sourceSets.functionalTest.java.srcDirs
+        testSourceDirs += project.sourceSets.functionalTest.resources.srcDirs
+        testSourceDirs += project.sourceSets.smokeTest.java.srcDirs
+        testSourceDirs += project.sourceSets.smokeTest.resources.srcDirs
+    }
```
t80: Should the resource directories be specified as testResourceDirs

src/main/resources/db/migration/V1.0.1__init_tables.sql
```diff
@@ -0,0 +1,12 @@
+--
+CREATE SCHEMA IF NOT EXISTS ia_case_api;
+--
+CREATE TABLE IF NOT EXISTS ia_case_api.appeal_reference_numbers (
+  case_id INT NOT NULL,
```
t80: I thought this was changed to Long or BigInt?

t80: Ignore me

PR #95, opened 2019-05-13
Ria 941 generate hearing ready bundle

raghera:

https://tools.hmcts.net/jira/browse/RIA-941

### Change description ###

See Jira above.

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/util/SystemDocumentManagementUploader.java
```diff
@@ -0,0 +1,85 @@
...
+            UploadResponse uploadResponse =
+                documentUploadClientApi
+                    .upload(
+                        accessToken,
+                        serviceAuthorizationToken,
+                        userId,
+                        Collections.singletonList(file)
+                    );
+
+            uk.gov.hmcts.reform.document.domain.Document uploadedDocument =
```
t80: Is there another Class name that would make sense to avoid fully qualified class name?

raghera: This is from a library we use from `document-management-client` which clashed with our own `Document` object name.  I think since it's just in one place I think it's acceptable.  Later we can consider refactoring our internal `Document` to be named something like `CaseDocument` or similar.

raghera:
Merged as part of 1492 as a feature flag was required around this functionality.

PR #104, opened 2019-06-10
RIA-1539 pre-submit and post-submit callback handlers ordering

makowalski:
[RIA-1539](https://tools.hmcts.net/jira/browse/RIA-1539)

### Change description ###
pre-submit callback handlers are sorted based on dispatch priority and handler name (class name) value

Spring doesn't guarantee the same order of beans in injected list. There is a way to force Spring to use desired order.
[https://www.baeldung.com/spring-injecting-collections](https://www.baeldung.com/spring-injecting-collections) 5.1

I think the best way is sorting handlers in constructor than thinking about another annotation in handler bean component.

```
[ ] Yes
[x] No
```


makowalski:
Guys, if you are happy with solution I will create pull requests for ia-case-documents-api and ia-case-notifications-api

PR #135, opened 2019-08-23
Ria 1380 add hearing centres

bradsorour:

[RIA-1380](https://tools.hmcts.net/jira/browse/RIA-1380)

### Change description ###

See Jira

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/HearingCentreFinder.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/HearingCentreFinder.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.reform.iacaseapi.domain.service;
 
 import com.google.common.collect.ImmutableMap;
+import java.time.LocalDate;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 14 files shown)


src/main/resources/application-postcodes.yaml
```diff
@@ -0,0 +1,78 @@
+hearingCentreActivationDates:
```
raghera: Good that this is in it's own file now.

src/main/resources/application.yaml
```diff
@@ -37,6 +37,9 @@ spring:
       remove-abandoned-timeout: 60
       log-abandoned: true
       abandon-when-percentage-full: 0
+  profiles:
+    include:
+      - postcodes
```
raghera: is this to add the new yaml file?

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/HearingCentreFinder.java
```diff
@@ -46,10 +50,28 @@ public HearingCentre find(
...
+                if (hearingCentreIsActive(hearingCentreActivationDate)) {
+                    return hearingCentre.get();
+                }
             }
         }
 
         return defaultHearingCentre;
     }
+
+    public boolean hearingCentreIsActive(String hearingCentreActivationDate) {
```
raghera: Do we care about the time here?  Does the activation for instance intended to happen after 00:00 on the configured date or the next day?

src/main/resources/application.yaml
bradsorour: yes

PR #141, opened 2019-09-02
RIA-1562 Launch end appeal from overview tab

tarun-palisetty:
[RIA-1562](https://tools.hmcts.net/jira/browse/RIA-1562)

### Change description ###
Launch end appeal from application

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -232,6 +232,11 @@ dependencyManagement {
             entry 'tomcat-embed-el'
             entry 'tomcat-embed-websocket'
         }
+
+        // CVE-2019-10086
+        dependencySet(group: 'commons-beanutils', version: '1.9.4') {
```
makowalski: it is currently fixed in master, we had to upgrade checkstyle plugin version to use the latest commons-beanutils
please remove that set, I think it is not needed now

PR #143, opened 2019-09-05
Ria 1304 authenticate home office user

bradsorour:

[RIA-1304](https://tools.hmcts.net/jira/browse/RIA-1304)

### Change description ###

See Jira

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -129,6 +129,31 @@ security:
       - "listCase"
       - "recordAttendeesAndDuration"
       - "editCaseListing"
+    caseworker-ia-homeofficeapc:
+      - "uploadRespondentEvidence"
+    caseworker-ia-homeofficelart:
+      - "addAppealResponse"
+    caseworker-ia-homeofficepou:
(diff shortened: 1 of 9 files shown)


src/main/resources/application.yaml
```diff
@@ -127,6 +127,14 @@ security:
...
       - "listCase"
       - "recordAttendeesAndDuration"
+    caseworker-ia-homeofficeapc:
+      - "uploadRespondentEvidence"
+    caseworker-ia-homeofficelart:
+      - "addAppealResponse"
+    caseworker-ia-homeofficepou:
+      - "sendDirection"
+    caseworker-ia-respondentofficer:
+      - "sendDirection"
```
jkumar-moj: My understanding is respondent office should be able to do everything that other homeoffice roles could do. Eg: addAppealResponse and uploadRespondentEvidence. Could you pleae clarify?

bradsorour: I think you're correct, thanks @jkumar-moj

PR #163, opened 2019-10-07
Enable AKS in AAT for ia-case-api

raghera:
Enable AKS in AAT for ia-case-api.


raghera:
Already been done in another PR

PR #180, opened 2019-11-11
performance logs

makowalski:
N/A

### Change description ###
performance logs

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/CcdEventAuthorizor.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/CcdEventAuthorizor.java
@@ -1,19 +1,17 @@
 package uk.gov.hmcts.reform.iacaseapi.infrastructure.security;
 
-import static org.slf4j.LoggerFactory.getLogger;
-
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.security.access.AccessDeniedException;
 import uk.gov.hmcts.reform.iacaseapi.domain.UserDetailsProvider;
 import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.Event;
 
+@Slf4j
 public class CcdEventAuthorizor {
-
-    private static final org.slf4j.Logger log = getLogger(CcdEventAuthorizor.class);
 
     private final Map<String, List<Event>> roleEventAccess;
     private final UserDetailsProvider userDetailsProvider;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/CcdEventAuthorizor.java
```diff
@@ -1,15 +1,20 @@
...
 import java.util.Map;
 import java.util.stream.Collectors;
+
 import org.springframework.security.access.AccessDeniedException;
 import uk.gov.hmcts.reform.iacaseapi.domain.UserDetailsProvider;
 import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.Event;
 
 public class CcdEventAuthorizor {
 
+    private static final org.slf4j.Logger log = getLogger(CcdEventAuthorizor.class);
```
raghera: Can use the @Slf4j annotation from Lombok which saves you from declaring the logger

jkumar-moj: Could we keep the variable to capitals to keep it consistent? LOG instead of log?

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/idam/IdamUserDetailsProvider.java
```diff
@@ -17,6 +19,8 @@
 
 public class IdamUserDetailsProvider implements UserDetailsProvider {
 
+    private static final org.slf4j.Logger log = getLogger(IdamUserDetailsProvider.class);
```
jkumar-moj: Same as above

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/CcdEventAuthorizor.java
```diff
@@ -24,6 +29,8 @@ public CcdEventAuthorizor(
     public void throwIfNotAuthorized(
         Event event
     ) {
+        long startTime = System.currentTimeMillis();
```
raghera: This is fine for what we want here.  In future we could consider something like:  https://commons.apache.org/proper/commons-lang/javadocs/api-3.9/org/apache/commons/lang3/time/StopWatch.html

Don't think we need it here though.

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/CcdEventAuthorizor.java
makowalski: Ok, let me change all places in ia-case-api for being consistent (please see @jkumar-moj comment)

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/security/CcdEventAuthorizor.java
makowalski: I am thinking of generic IA http client (RestTemplate wrapper) for 3rd party services with proper response logging (in case of exception) and optional timing (env flag), it would be more elegant. What do you think?

raghera: I think some generic standardised logging for our interactions with other services would be good.  I think I prefer it being separate from the RestTemplate however.  Maybe we can re-use what Spring gives us with @ControllerAdvice and add some generic logging there?

PR #205, opened 2019-11-21
 RIA-606: Added validations for Home Office decision date and appella…

jkumar-moj:
…nt date of birth during appeal start and edit
https://tools.hmcts.net/jira/browse/RIA-606

### Change description ###
Home Office decision date and appellant DOB validation

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/CaseDetails.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/CaseDetails.java
@@ -47,10 +47,6 @@ public class CaseDetails<T extends CaseData> {
 
     public State getState() {
 
-        if (state == null) {
-            throw new RequiredFieldMissingException("state field is required");
-        }
-
         return state;
     }
 
@@ -64,10 +60,6 @@ public class CaseDetails<T extends CaseData> {
     }
 
     public LocalDateTime getCreatedDate() {
-
-        if (createdDate == null) {
-            throw new RequiredFieldMissingException("createdDate field is required");
-        }
 
         return createdDate;
     }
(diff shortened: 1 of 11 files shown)


charts/ia-case-api/values.preview.template.yaml
```diff
@@ -31,6 +31,6 @@ java:
         - docmosis-enabled
         - em-stitching-enabled
 
-postgresql:
-  #Whether to deploy the Postgres Chart or not
-  enabled: true
\ No newline at end of file
+  postgresql:
```
raghera: Did you mean to change the formatting here?

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/CaseDetails.java
```diff
@@ -47,10 +47,6 @@ public String getJurisdiction() {
 
     public State getState() {
 
-        if (state == null) {
```
raghera: Why did we remove these?

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/controllers/PreSubmitCallbackController.java
```diff
@@ -124,6 +124,13 @@ public PreSubmitCallbackController(
         return performStageRequest(PreSubmitCallbackStage.ABOUT_TO_SUBMIT, callback);
     }
 
+    @PostMapping(path = "/ccdMidEvent")
```
raghera: We should add swagger api docs for this new method.  Happy to do that as a separate task but it should be there since we are going to public beta very soon.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/ccd/CaseDetails.java
jkumar-moj: This validation happens before ccd assigns state to a case and hence this condition had to be taken out. Discussed this with @Marcin and we checked with you as well during implementation.

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/controllers/PreSubmitCallbackController.java
jkumar-moj: Thanks ravi. I will add the documentation.

PR #214, opened 2019-11-29
fix to enable legacy deployment

bradsorour:
Fix required to use ase not aks in Demo

```
[ ] Yes
[x] No
```


raghera:
Closing as don't want an open PR on demo.  Will create another PR for it.

PR #230, opened 2019-12-17
Share a case core

makowalski:
[RIA-723](https://tools.hmcts.net/jira/browse/RIA-723)

### Change description ###
partially reverting changes for Direction object

```
[ ] Yes
[x] No
```


charts/ia-case-api/values.yaml
```diff
@@ -7,6 +7,7 @@ java:
     POSTGRES_HOST: ia-case-api-postgres-db-{{ .Values.global.environment }}.postgres.database.azure.com
     POSTGRES_USERNAME: ia_case_api@ia-case-api-postgres-db-{{ .Values.global.environment }}
     POSTGRES_CONNECTION_OPTIONS: "?sslmode=require"
+    PROF_REF_DATA_URL: "http://rd-professional-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
```
raghera: If we do this ... in theory we would not need the values in the vault as it sets the env correctly from global vars.

src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/CcdScenarioRunnerTest.java
```diff
@@ -134,6 +134,8 @@ public void scenarios_should_behave_as_specified() throws IOException {
             final String requestUri = MapValueExtractor.extract(scenario, "request.uri");
             final int expectedStatus = MapValueExtractor.extractOrDefault(scenario, "expectation.status", 200);
 
+            System.out.println();
```
raghera: can probably get rid of this

src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/ShareACaseScenarioTest.java
```diff
@@ -0,0 +1,48 @@
+package uk.gov.hmcts.reform.iacaseapi;
```
raghera: This file is not required.

src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/CcdScenarioRunnerTest.java
makowalski: :+1:

src/functionalTest/java/uk/gov/hmcts/reform/iacaseapi/ShareACaseScenarioTest.java
makowalski: :+1:

charts/ia-case-api/values.yaml
makowalski: as discussed, there is a need to refactor this and other urls which shouldn't be stored in Vault

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/ShareACaseUserListPreparer.java
```diff
@@ -0,0 +1,80 @@
...
+
+        AsylumCase asylumCase = callback.getCaseDetails().getCaseData();
+
+        DynamicList dynamicList;
+
+        final List<Value> values =
+            usersResponse.getProfessionalUsers()
+                .stream()
+                .filter(professionalUser ->
+                    professionalUser.getIdamStatus().equalsIgnoreCase("ACTIVE"))
```
raghera: In future we can use the parameter in PRD API that allows a filtered response.  This is good for the core version

makowalski: :+1:

PR #234, opened 2020-01-03
AKS migration for ITHC env

makowalski:
N/A

### Change description ###
AKS migration for ITHC env

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/ithc.tfvars
+++ b/infrastructure/ithc.tfvars
@@ -1,0 +1,3 @@
+capacity = "1"
+instance_size="I1"
+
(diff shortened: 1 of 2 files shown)


infrastructure/ithc.tfvars
```diff
@@ -0,0 +1,3 @@
+capacity = "1"
```
raghera: Do we need this if we are just using AKS?

makowalski: Probably we do not need, but we have the files for all envs for now. I didn't want to make exception for ithc

makowalski: I confirmed with Praveen, we still need this file because there are references in main.tf for these two variables.
We can't remove them from main.tf for now because these values are used in Demo env (it is still in ASE)

raghera: 👍

PR #259, opened 2020-01-21
RIA-2373 grant update hearing requirements

tarun-palisetty:
[RIA-2373](https://tools.hmcts.net/jira/browse/RIA-2373)

### Change description ###
As a case officer record granted update hearing requirements.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
@@ -385,6 +385,9 @@ public enum AsylumCaseFieldDefinition {
     SUBMIT_HEARING_REQUIREMENTS_AVAILABLE(
         "submitHearingRequirementsAvailable", new TypeReference<YesOrNo>() {}),
 
+    APPLICATION_UPDATE_HEARING_REQUIREMENTS_EXISTS(
+        "applicationUpdateHearingRequirementsExists", new TypeReference<String>() {})
+
     ;
 
     private final String value;
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/RecordApplicationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/RecordApplicationHandler.java
@@ -46,6 +46,14 @@ public class RecordApplicationHandler implements PreSubmitCallbackHandler<Asylum
         State.SUBMIT_HEARING_REQUIREMENTS
     );
 
+    private final List<State> updateHearingRequirementsStates = newArrayList(
+        State.PRE_HEARING,
+        State.FINAL_BUNDLING,
+        State.PREPARE_FOR_HEARING,
+        State.DECISION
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/RecordApplicationHandler.java
```diff
@@ -156,6 +156,10 @@ public boolean canHandle(
                 if (!asylumCase.read(APPLICATION_WITHDRAW_EXISTS, String.class).isPresent()) {
                     asylumCase.write(APPLICATION_EDIT_LISTING_EXISTS, "Yes");
                 }
+            } else if (UPDATE_HEARING_REQUIREMENTS.toString().equalsIgnoreCase(applicationType)) {
+                if (!asylumCase.read(UPDATE_HEARING_REQUIREMENTS_EXISTS, String.class).isPresent()) {
+                    asylumCase.write(UPDATE_HEARING_REQUIREMENTS_EXISTS, "Yes");
+                }
             }
 
             asylumCase.write(DISABLE_OVERVIEW_PAGE, "Yes");
```
jkumar-moj: I think clearing the flag needs to be handled for this event in an appropriate place. I think it is cleared specifically based on 'APPLICATION_EDIT_LISTING_EXISTS' flag which has no link to this event. Not sure to be handled as a part of this change or some follow up change. Just flagging.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
```diff
@@ -377,7 +377,10 @@
         "interpreterLanguageReadonly", new TypeReference<List<IdValue<InterpreterLanguage>>>() {}),
 
     SUBMIT_HEARING_REQUIREMENTS_AVAILABLE(
-        "submitHearingRequirementsAvailable", new TypeReference<YesOrNo>() {})
+        "submitHearingRequirementsAvailable", new TypeReference<YesOrNo>() {}),
+
+    UPDATE_HEARING_REQUIREMENTS_EXISTS(
+        "updateHearingRequirementsExists", new TypeReference<String>() {})
```
makowalski: It would be good to have prefix "application" for the flag. There is for all Application related flags. It is minor thing but still helpful to manage all those flags.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/RecordApplicationHandler.java
makowalski: Clearing the flag will be in my ticket: RIA-2031. At some point we need to merge everything and deploy in one step to avoid broken links

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
tarun-palisetty: prefixed the flag with 'application'

PR #276, opened 2020-02-06
Ria 2541 reasons for appeal enter data

chrisgrimble:
When a case office requests the reasons for appeal they are now
presented with a form to select the required by date. Added code to
populate the form and set the date to 4 weeks in the future. Made
REQUEST_RESONS_FOR_APPEAL a direction so it will be seen in the
directions tab.

https://tools.hmcts.net/jira/browse/RIA-2541


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/Parties.java
```diff
@@ -6,7 +6,8 @@
 
     LEGAL_REPRESENTATIVE("legalRepresentative"),
     RESPONDENT("respondent"),
-    BOTH("both");
+    BOTH("both"),
+    APPELLANT("appellant");
```
makowalski: @chrisgrimble Adding something to the domain enum is a bit complicated because of CCD defs and other services.
There is a need to add this change to notifications and documents first otherwise jackson parser will fail and eventually the event callback too.

Do you plan to add "appellant" option to CCD defs (FixedLists -> parties)? If CCD sees something which is not in the definitions returns validation error for the case.

If you add this to CCD defs, everyone will see that option in sendDirection event for case officer, you will be able to send direction to the appellant even if the case is REP.

chrisgrimble: This was added to the the CCD def yes. If it was intended to allow send directions to an appellant I just did what I was told.

PR #286, opened 2020-02-13
RIA-2632 Prevent send direction for AIP journey

chrisgrimble:
Prevent the send direction event being fired for an aip journey as we
have not implemented this yet.
Prevent caseworker selecting Appellant for send direction for a repped
case.

https://tools.hmcts.net/jira/browse/RIA-2632


src/test/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/eventvalidation/AsylumCaseEventValidForJourneyTypeCheckerTest.java
```diff
@@ -49,7 +49,7 @@ public void canSendValidEvents() {
     public void cannotSendInvalidEventsForAip() {
         setupCallback(Event.REQUEST_CASE_BUILDING, AIP);
 
-        EventValidForJourneyType check = asylumCaseEventValidForJourneyTypeChecker.check(callback);
+        EventValid check = asylumCaseEventValidForJourneyTypeChecker.check(callback);
```
raghera: should we add a case for sendDirection specifically?

PR #313, opened 2020-03-09
Ria 1293 ftpa appellant

jkumar-moj:
https://tools.hmcts.net/jira/browse/RIA-1293

### Change description ###
FTPA Appellant application to the Upper Tribunal

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -91,6 +91,7 @@ cache:
     expirationTimeInSeconds: 60
 
 appealOutOfTimeDays: 14
+ftpaAppealOutOfTimeDays: 14
 
 defaultHearingCentre: taylorHouse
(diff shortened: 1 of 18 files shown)


src/main/resources/application.yaml
```diff
@@ -182,6 +186,7 @@ security:
     caseworker-ia-iacjudge:
       - "generateDecisionAndReasons"
       - "sendDecisionAndReasons"
+      - "applyForFTPAAppellant"
```
makowalski: I think only legal rep should have possibility to apply for FTPL and Home Office within next ticket. What do you think?

jkumar-moj: Thanks @makowalski This has been fixed now.

PR #353, opened 2020-04-06
ia-case-api CCD fields validation

makowalski:
N/A

### Change description ###
ia-case-api CCD fields validation

```
[ ] Yes
[x] No
```

### DIFF
--- a/bin/validate_case_api.js
+++ b/bin/validate_case_api.js
@@ -1,0 +1,112 @@
+const process = require('process');
+const fs = require('fs');
+const walkSync = require('walk-sync');
+const request = require('request');
+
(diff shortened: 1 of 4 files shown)


bin/validate_case_api.js
```diff
@@ -0,0 +1,112 @@
+const process = require('process');
+const fs = require('fs');
+const walkSync = require('walk-sync');
+const request = require('request');
+
+const BASE_IA_CCD_DIR = process.env.IA_CCD_DIR !== undefined ? process.env.IA_CCD_DIR : '../ia-ccd-definitions/';
```
makowalski, line 6: I will add remote reading for CCD definitions as default and leave local directory reading as optional within next PR.

PR #355, opened 2020-04-07
Ria 1432 notify ftpa appeal decision

burale:

[RIA-1432](https://tools.hmcts.net/jira/browse/RIA-1432)

### Change description ###

```
[ ] Yes
[x] No
```


burale:
It is expected to fail on functional test - AKS preview, because it depends on notification 1432 which is in prod yet.

PR #407, opened 2020-05-19
RIA-2952 upgrading to junit 5

bradsorour:

[RIA-2952](https://tools.hmcts.net/jira/browse/RIA-2952)

### Change description ###

See Jira

```
[ ] Yes
[x] No
```


bradsorour:
@raghera Yes, the JUnit 4 and 5 tests run side by side for this PR. I have, in fact, stashed changes on this branch where I updated all the tests. As you pointed out, this was straight-forward and quick to do using find/replace. The only unit test requiring additional attention is the EditDocsAuditLogHandlerTest, which needs its annotation replaced with @ParameterizedTest and an Arguments Provider. This change is not needed in the other two services, namely: ia-case-notifications-api and ia-case-documents-api.

PR #427, opened 2020-06-03
Ria 2466 fee hearing option

Lee365:
https://tools.hmcts.net/jira/browse/RIA-2466

### Change description ###
RIA-2466 Fee Hearing Option 

```
[ ] Yes
[X] No
```


charts/ia-case-api/values.yaml
```diff
@@ -42,3 +43,4 @@ java:
         - launch-darkly-sdk-key
         - case-api-POSTGRES-PASS
         - AppInsightsInstrumentationKey
+        - launch-darkly-sdk-key
```
makowalski, line 46: this key is already in master

build.gradle
```diff
@@ -289,6 +289,7 @@ dependencies {
 
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
     compile group: 'uk.gov.service.notify', name: 'notifications-java-client', version: '3.12.0-RELEASE'
+    compile group: 'com.launchdarkly', name: 'launchdarkly-java-server-sdk', version: '4.12.1'
```
makowalski, line 292: it is there already: https://github.com/hmcts/ia-case-api/blob/master/build.gradle#L298

charts/ia-case-api/values.yaml
Lee365: Removed

build.gradle
Lee365: Removed

PR #513, opened 2020-08-11
Ria 2923 customise bundle

mkayad:

https://tools.hmcts.net/jira/browse/RIA-2923

### Change description ###

add hearing bundle customisation

```
[ ] Yes
[ X] No
```


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/CustomiseHearingBundleHandler.java
```diff
@@ -0,0 +1,281 @@
...
+        asylumCase.clear(AsylumCaseFieldDefinition.CASE_BUNDLES);
+        asylumCase.write(AsylumCaseFieldDefinition.BUNDLE_CONFIGURATION, "iac-hearing-bundle-config.yaml");
+        asylumCase.write(AsylumCaseFieldDefinition.BUNDLE_FILE_NAME_PREFIX, getBundlePrefix(asylumCase));
+
+        final PreSubmitCallbackResponse<AsylumCase> response = bundleRequestExecutor.post(callback, emBundlerUrl + emBundlerStitchUri);
+
+        final AsylumCase responseData = response.getData();
+
+        //Customise the folders before returning
+        CaseDetails<AsylumCase> caseDetailsBefore = callback.getCaseDetailsBefore().orElse(null);
```
makowalski, line 90: We can't use caseDetailsBefore from callback. It is not the version one event before our event. It is like reference to previous, but we need to have reference to current.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/CustomiseHearingBundleHandler.java
```diff
@@ -0,0 +1,281 @@
...
+        if (!canHandle(callbackStage, callback)) {
+            throw new IllegalStateException("Cannot handle callback");
+        }
+
+        final AsylumCase asylumCase =
+                callback
+                        .getCaseDetails()
+                        .getCaseData();
+
+        prepareDocuments(asylumCase, CUSTOM_HEARING_DOCUMENTS,HEARING_DOCUMENTS);
```
makowalski, line 73: I would rather create deep-copy of `asylumCase` for example `asylumCaseCopy` by using Jackson mapping example below:
https://www.baeldung.com/java-deep-copy
and pass to `prepareDocuments` method

The original `asylumCase` could be used as a return response object after adding `CASE_BUNDLES` and `STITCHING_STATUS` fields

We can't use caseDetailsBefore for that purpose.

PR #752, opened 2021-01-25
RWA-319 point role-assignment to ia-docker wiremock by default

iguannago:
https://tools.hmcts.net/jira/browse/RWA-319


makowalski (approved):
LGTM

PR #764, opened 2021-02-02
RWA-275 add wa role

iguannago:
https://tools.hmcts.net/jira/browse/RWA-275


lewis150193 (approved):

alexbaeza (approved):

iguannago:
It's no longer needed because we can use the legal role and legal from our tests to read cases from CCD.

PR #771, opened 2021-02-09
RIA-3990 case-details class updated with case-type-id field

tarun-palisetty:
[RIA-3990](https://tools.hmcts.net/jira/browse/RIA-3990)

### Change description ###
Case-data updated with case-type-id

```
[ ] Yes
[x] No
```


makowalski (dismissed):
LGTM

PR #783, opened 2021-02-11
Additional id logging for different PRD endpoints

bradsorour:

N/A

### Change description ###

Additional ID logging for our PRD endpoint call. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/LegalRepOrganisationFormatter.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/LegalRepOrganisationFormatter.java
@@ -63,6 +63,10 @@ public class LegalRepOrganisationFormatter implements PreSubmitCallbackHandler<A
         if (organisationEntityResponse != null
             && StringUtils.isNotBlank(organisationEntityResponse.getOrganisationIdentifier())
             && featureToggler.getValue("share-case-feature", false)) {
+
+            log.info("PRD endpoint called for caseId [{}] orgId[{}]",
+                callback.getCaseDetails().getId(), organisationEntityResponse.getOrganisationIdentifier());
+
             setupCaseCreation(callback, organisationEntityResponse.getOrganisationIdentifier());
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/clients/ProfessionalOrganisationRetriever.java
```diff
@@ -73,6 +74,11 @@ public OrganisationEntityResponse retrieve() {
 
         log.info("response returned: {}", response);
 
+        log.info(
+            "PRD endpoint {} : userId[{}] orgId[{}]",
+            refDataApiPath, userDetails.getId(), Objects.requireNonNull(response).getOrganisationIdentifier()
```
makowalski, line 79: this may throw NPE when response is null. Returning null at this point is ok. 
You can add userId to the existing log (line 75) and add Lombok `@ToString` annotation to `OrganisationEntityResponse`  so the existing log would be ok

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/LegalRepOrganisationFormatter.java
```diff
@@ -63,6 +63,9 @@ public boolean canHandle(PreSubmitCallbackStage callbackStage, Callback<AsylumCa
         if (organisationEntityResponse != null
             && StringUtils.isNotBlank(organisationEntityResponse.getOrganisationIdentifier())
             && featureToggler.getValue("share-case-feature", false)) {
+
+            log.info("PRD endpoint called for caseId [{}] ", callback.getCaseDetails().getId());
```
makowalski, line 67: here we can add OrgId we know it is not blank at this point

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/clients/ProfessionalOrganisationRetriever.java
bradsorour: Done, thanks.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/LegalRepOrganisationFormatter.java
bradsorour: Done.

jkumar-moj (approved):

PR #809, opened 2021-03-02
Task/allocate the case fix

iguannago:
rwa-318 and rwa-319, hystrix fix.


src/integrationTest/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/allocatethecase/CaseWorkerServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+import uk.gov.hmcts.reform.iacaseapi.domain.handlers.presubmit.allocatecase.CaseWorkerName;
+import uk.gov.hmcts.reform.iacaseapi.domain.service.CaseWorkerService;
+import uk.gov.hmcts.reform.iacaseapi.domain.service.IdamService;
+
+@SpringBootTest
+@ActiveProfiles("integration")
+public class CaseWorkerServiceTest {
+
+    @Autowired
+    private WireMockServer caseWorkerRefDataServiceMock;
```
makowalski, line 22: There is common way of using wiremock in inregration tests in case-api, it would be good to keep it consistent if in case we need some refactoring, with JUnit 5 it can be injected with annotation:
```
public void adds_a_case_note(
        @WiremockResolver.Wiremock(factory = StaticPortWiremockFactory.class) WireMockServer server)
```

src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/config/DisableHystrixFeignConfiguration.java
```diff
@@ -2,10 +2,8 @@
 
 import feign.Feign;
 import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Scope;
 
-@Configuration
 public class DisableHystrixFeignConfiguration {
```
makowalski, line 7: if I understand correctly:
removing annotation gives you flexibility to inject configuration to particular feign client, I guess Configuration files are scanned at the startup and that could be injected to case worker feign client

makowalski (approved):
LGTM

PR #836, opened 2021-03-23
RIA-3996

Lee365:
https://tools.hmcts.net/jira/browse/RIA-3996

### Change description ###
RIA-3996 CCD reference visible to admin

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
@@ -1400,6 +1400,8 @@ public enum AsylumCaseFieldDefinition {
     DEPORTATION_ORDER_OPTIONS(
         "deportationOrderOptions", new TypeReference<YesOrNo>(){}),
 
+    CCD_REFERENCE_NUMBER_FOR_DISPLAY(
+        "ccdReferenceNumberForDisplay", new TypeReference<String>(){}),
     ;
 
     private final String value;
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AsylumCaseCcdReferenceNumberHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AsylumCaseCcdReferenceNumberHandler.java
@@ -1,0 +1,57 @@
+package uk.gov.hmcts.reform.iacaseapi.domain.handlers.presubmit;
+
+import static java.util.Objects.requireNonNull;
+import static uk.gov.hmcts.reform.iacaseapi.domain.entities.AsylumCaseFieldDefinition.CCD_REFERENCE_NUMBER_FOR_DISPLAY;
+
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.AsylumCase;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.CaseDetails;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.Callback;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.DispatchPriority;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.PreSubmitCallbackResponse;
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AppealReferenceNumberHandler.java
```diff
@@ -69,6 +69,11 @@ public boolean canHandle(
 
         if (callback.getEvent() == Event.START_APPEAL) {
             asylumCase.write(APPEAL_REFERENCE_NUMBER, DRAFT);
+
+            String caseId = "" + callback.getCaseDetails().getId();
```
bradsorour, line 73: Would `String caseId = String.valueOf(callback.getCaseDetails().getId())` not work in this case. 

You may not need to run `replaceAll` on it afterwards.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/AsylumCaseFieldDefinition.java
```diff
@@ -1400,6 +1400,8 @@
     DEPORTATION_ORDER_OPTIONS(
         "deportationOrderOptions", new TypeReference<YesOrNo>(){}),
 
+    CCD_REFERENCE_NUMBER_FOR_DISPLAY(
```
bradsorour, line 1403: I would consider calling the `CCD_CASE_ID` or `CCD_CASE_ID_FOR_DISPLAY` for better accuracy.

Lee365: Left as CCD_REFERENCE_NUMBER_FOR_DISPLAY

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AppealReferenceNumberHandler.java
Lee365: Changed to String caseId = String.valueOf(callback.getCaseDetails().getId());

bradsorour (dismissed):
LGTM

Srihari-Nakkala (approved):
LGTM

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AppealReferenceNumberHandler.java
```diff
@@ -69,6 +69,12 @@ public boolean canHandle(
 
         if (callback.getEvent() == Event.START_APPEAL) {
             asylumCase.write(APPEAL_REFERENCE_NUMBER, DRAFT);
+
+            String caseId = String.valueOf(callback.getCaseDetails().getId());
+            caseId = caseId.replaceAll("(.{" + 4 + "})", "$1 ").trim();
+
+            asylumCase.write(CCD_REFERENCE_NUMBER_FOR_DISPLAY,  caseId);
+
```
bradsorour, line 77: @Lee could you look to move this code into a new handler (i.e. a handler that will be triggered for all events) so the CCD_REFERENCE_NUMBER_FOR_DISPLAY is the available for all appeals.

If you take a look at the `AsylumCaseDataFixingHandler` class, you will see that the `canHandle()` method always returns true - so a handler similar to this. 

If we have the code in the `AppealReferenceNumberHandler`, the CCD_REFERENCE_NUMBER_FOR_DISPLAY will only display/be written to the AsylumCase for new cases.

bradsorour:
@Lee365 a change to this PR if you can please (identified when planning NoC deployment with @makowalski)

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AppealReferenceNumberHandler.java
makowalski: Yes, I agree. I think the cleanest way is to use new dedicated handler for it.
There is an option to use fixers but the whole part of the fixers should be refactored before they are more useful.

Lee365: Changed to use new handler - AsylumCaseCcdReferenceNumberHandler

makowalski (approved):
LGTM

PR #887, opened 2021-05-13
RIA-4369 initialize appealOutOfCountry case-field for the older appeals

tarun-palisetty:
[RIA-4369](https://tools.hmcts.net/jira/browse/RIA-4363)

### Change description ###
appealOfOutCountry case-field is being initialised for older appeals.
As appealOutOfCountry case-field is missing for the older appeals causing submit-hearing-requirements to fail as it is a mandatory field in the event.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/DraftHearingRequirementsHandlerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/DraftHearingRequirementsHandlerTest.java
@@ -10,17 +10,7 @@ import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/DraftHearingRequirementsHandlerTest.java
```diff
@@ -131,6 +133,7 @@ void should_set_witness_count_and_available_fields() {
 
         when(asylumCase.read(WITNESS_DETAILS)).thenReturn(Optional.of(Arrays
             .asList(new IdValue("1", new WitnessDetails("cap")), new IdValue("2", new WitnessDetails("Pan")))));
+        when(asylumCase.read(APPEAL_OUT_OF_COUNTRY, YesOrNo.class)).thenReturn(Optional.of(YesOrNo.YES));
```
Lee365, line 136: APPEAL_OUT_OF_COUNTRY is not available and should return empty. This can be then checked to see if its saved with default 'NO' value.

Lee365 (changes requested):

Lee365 (approved):
LGTM

PR #917, opened 2021-06-22
RWA-584 Add attribute caseManagementCategory to IAC CCD definition

mkayad:

[RWA-584](https://tools.hmcts.net/jira/browse/RWA-584)

### Change description ###

See ticket

```
[ ] Yes
[X ] No
```


iguannago (approved):

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/CaseManagementCategoryAppenderTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        assertThat(response).isNotNull();
+        assertThat(response.getData()).isNotEmpty();
+        assertThat(response.getData()).isEqualTo(asylumCase);
+        assertThat(response.getErrors()).isEmpty();
+
+
+        Mockito.verify(asylumCase, times(1)).write(CASE_MANAGEMENT_CATEGORY, dynamicList);
+    }
+
```
BenNzewi, line 65: should the test also check the other possible values as per AC? I.e.
refusalofHumanRights
refusalOfEU
deprivation
revocationOfProtection

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/CaseManagementCategoryAppender.java
```diff
@@ -0,0 +1,59 @@
...
+            PreSubmitCallbackStage callbackStage,
+            Callback<AsylumCase> callback
+    ) {
+        if (!canHandle(callbackStage, callback)) {
+            throw new IllegalStateException("Cannot handle callback");
+        }
+
+        AsylumCase asylumCase = callback.getCaseDetails().getCaseData();
+
+        Optional<AppealType> appealType
```
raghera, line 48: Is an appealType a `must` in a case?  In which case if it's not present I would suggest throwing an illegal state exception.  @makowalski what do you think?

raghera: We also need to think about historic case data that will not have the new field present.  We have a pattern whereby we can fill these in on callbacks such as here: https://github.com/hmcts/ia-case-api/blob/master/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/AddBaseLocationFromHearingCentreForOldCasesFixHandler.java

The only gap then will be cases where no CCD Events have run successfully since we implement the above

raghera (dismissed):

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/CaseManagementCategoryAppenderTest.java
mkayad: test improved

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/CaseManagementCategoryAppender.java
makowalski: @raghera @mkayad 
appelType is mandatory field on startAppel event (first event in the appeal lifecycle). As Ravi suggested we should use Illegal State Exception to handle missing appealType field. We are doing it in several places in code already.
```
  {
    "LiveFrom": "01/01/2018",
    "CaseTypeID": "Asylum",
    "CaseEventID": "startAppeal",
    "CaseFieldID": "appealType",
    "CaseEventFieldLabel": "Type of appeal",
    "PageFieldDisplayOrder": 2,
    "DisplayContext": "MANDATORY",
    "PageID": "appealType",
    "PageDisplayOrder": 18,
    "ShowSummaryChangeOption": "Y"
  },
```

mkayad: Done

mkayad:
Issue with FT in AKS env

indvamshi (approved):

PR #932, opened 2021-07-08
https://tools.hmcts.net/jira/browse/RWA-640

iguannago:
https://tools.hmcts.net/jira/browse/RWA-640

fix the wa dev env


bradsorour (approved):
LGTM

PR #971, opened 2021-09-02
RIA-4691: Removed edit appeal after submit validation for AiP

jkumar-moj:
https://tools.hmcts.net/jira/browse/RIA-4691

### Change description ###
Removed edit appeal after submit validation for AiP

```
[ ] Yes
[X] No
```


jkumar-moj:
This is no more valid. This was added initially as editAppealAfterSubmit was used in the develoment. Since that is not the case we will stil need this validation for AIP. Closing the PR

PR #1017, opened 2022-01-13
RIA-4542 set Newport as the default Hearing Center

augustio:
Please remove this line and everything above and fill the following sections:

https://tools.hmcts.net/jira/browse/RIA-4542

### Change description ###

```
[ ] Yes
[x] No
```


src/functionalTest/resources/scenarios/RIA-2323-edit-case-listing-docmosis-disabled-nottingham.json
```diff
@@ -104,7 +104,6 @@
           }
         ],
         "listCaseHearingCentre": "nottingham",
-        "hearingCentre": "taylorHouse",
```
bradsorour, line 107: @augustio do we need to delete this field, should this not be replaced with newport?

augustio: I could replace it with newport, however, I thought since the test expects the hearing centre to be the default hearing centre (newport) and the template used (minimal-appeal-submitted.js) is already set with the default hearing centre the default hearing centre, that there was no need of adding that replacement. What do you think?

bradsorour: @augustio okay yes. It's set in `minimal-appeal-submitted.json` so the removal makes sense.

bradsorour (approved):
LGTM

burale (approved):
LGTM

PR #1021, opened 2022-02-08
Revert "RIA-4542 set Newport as the default Hearing Center"

augustio:
Reverts hmcts/ia-case-api#1017
Temporarily reverting this merge because it introduces a subtle breaking change for the Work allocation team.


bradsorour (approved):
LGTM

PR #1056, opened 2022-07-15
Ria 5742 remove old way of paying

Marcellofabbri:
[RIA-5742](https://tools.hmcts.net/jira/browse/RIA-5742)

### Change description ###
- PAY_FOR_APPEAL, PAY_AND_SUBMIT_APPEAL, EDIT_PAYMENT_METHOD seem to be events used by Legal Rep only (judging by the authorizations in ccd definitions).
- All three methods have been removed, as well as their references across ia-case-api
- Tests and functional tests that were based on these three events have been deleted

- What should be left is the new way of paying (not through events, but through Service Request) and the PAYMENT_APPEAL event, which is used by AIP (and should be unreachable by REP)

```
[ ] Yes
[x] No
```


sharmad-v1 (approved):
Looks good

PR #1091, opened 2022-08-31
RIA-6143 party-name and role-on-case supplementary data

JamieHMCTS:
https://tools.hmcts.net/jira/browse/RIA-6143

### Change description ###
Add the  party-name and role-on-case values to the supplementary data field for cases.

```
[ ] Yes
[ x] No
```


sibuckley-v1 (changes requested):
Need to have a rethink on this one as there is data we need to write to CCD

sibuckley-v1 (approved):

PR #1150, opened 2022-11-14
RIA-6455: Added new hearing centre 'Harmondsworth'

neha-aggarw:
- Set default hearing centre as Harmondsworth for ADA cases.


colmlynch60 (approved):
LGTM @neha-aggarw ! I must get you to show me how the functional tests work

PR #1163, opened 2022-11-23
RIA-6427 age assessment part 1

Marcellofabbri:
[RIA-6427](https://tools.hmcts.net/jira/browse/RIA-6427)

### Related to ###
[ia-ccd-definitions/pull/834](https://github.com/hmcts/ia-ccd-definitions/pull/834)

### Change description ###
- Added DecisionLetterDateValidator to check the decision letter date is not in the future
- Added AppealTypeHandler to manage AppealType and AppealTypeForDisplay
  - AppealType has (HU, EA, PA, RP, DC, AG) 6 entries / AppealTypeForDisplay has (HU, EA, PA, RP, DC) 5 entries
  - AppealTypeForDisplay is what is displayed on the frontend (needs to match a FixedList field in ia-ccd-definition) in journeys that require the user to pick from the list of appeal types
  - AppealType isn't visible on the UI ever, and it includes all the 6 entries
  - In START_APPEAL and EDIT_APPEAL, at the ABOUT_TO_SUBMIT stage, the AppealTypeHandler sets the AppealType:
    - if `ageAssessment=Yes` it sets AppealType.AG
    - if `ageAssessment=No/null` then the AppealTypeForDisplay fixedList will show on the user journey, and AppealTypeHandler maps the selection (HU, EA, PA, RP, DC) to the matching entry in AppealType

This whole thing was devised because hiding one entry of a fixedList on the UI is not possible, but we're in a situation where we need to have 6 existing appealTypes, but only 5 are selectable from the FE, and one is settable from the backend. So we're showing a different list with 5 appealTypes, then map the value to the real AppealType list.

```
[ ] Yes
[🤞] No
```


JamieHMCTS (approved):
LGTM

colmlynch60 (approved):
LGTM

PR #1165, opened 2022-11-23
RIA-6516

markbrooks725:
https://tools.hmcts.net/jira/browse/RIA-6516

### Change description ###
Ensure that ccdSupplementaryUpdater does not fire if user is a citizen.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/iacaseapi/infrastructure/clients/CcdSupplementaryUpdater.java
```diff
@@ -53,6 +54,10 @@ public void setHmctsServiceIdSupplementary(final Callback<AsylumCase> callback)
         if (featureToggler.getValue("wa-R3-feature", false)) {
             requireNonNull(callback, "callback must not be null");
 
+            if (hasCitizenRole(userDetails.getRoles())) {
```
rajeshthuraiyur, line 57: Can we add journeyType as additional check to ensure update is excluded only AIP appeals with Citizen profiles?

PR #1178, opened 2022-12-01
Ria 6526 list case for ada

Marcellofabbri:
[RIA-6526](https://tools.hmcts.net/jira/browse/RIA-6526)

### Change description ###
- Added `ListCaseForAcceleratedDetainedAppeal` which mirrors `ListCase` and does everything `ListCase` does (but is triggered when the case is in `awaitingRespondentEvidence` if it's an accelerated detained appeal.

```
[ ] Yes
[x] No
```


Marcellofabbri:
Old approach, new one here: https://github.com/hmcts/ia-case-api/pull/1179

PR #1180, opened 2022-12-02
RIA-6547: Enabled the functional tests

neha-aggarw:
- RIA-515* failed after WA R2 merge; added a LD flag to exclude WA R3 feature while running this test.
- RIA-3680* - failed due to change in EditAppealAfterSubmit handler. It was changed to be triggered only at MidEvent with pageId & about_to_submit in Edit_Appeal_After_Submit. It no longer sets the field Submission_out_of_time as Yes/No at Start/Edit Appeal. This scenario is already covered in RIA-369.


rajeshthuraiyur (approved):
LGTM

PR #1212, opened 2022-12-21
RIA-6428: Fix the state of AG appeal after submission

neha-aggarw:
- Behaviour is same as Hu/Ea appeal types.
- state = PaymentPending.
- Fixed functional tests.

```
[ ] Yes
[x] No
```


Marcellofabbri:
There's another class that handles appeal types and fees where AG should probably be added alongside EA/HU/PA/EU: `FeesAndStatusCheckPreparer`.
But I think it's more for 'editAppeal' than 'submitAppeal'. Not sure if it's worth adding it as part of this PR

Marcellofabbri (approved):
LGTM

colmlynch60 (approved):
LGTM

PR #1216, opened 2023-01-05
Changed success page info for ADA cases

JamieHMCTS:
https://tools.hmcts.net/jira/browse/RIA-6527

### Change description ###
Added conditional logic to the confirmation page so that a unique message is displayed only for ADA cases.

**Found an content issue on the success page when testing [RIA-6495](https://tools.hmcts.net/jira/browse/RIA-6495), If it is Non ADA case then the Success page content should not change when we list the case, please check..**

```
[ ] Yes
[x ] No
```


src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/postsubmit/ListCaseConfirmationTest.java
```diff
@@ -33,41 +37,62 @@ class ListCaseConfirmationTest {
 
     private ListCaseConfirmation listCaseConfirmation = new ListCaseConfirmation();
 
-    @Test
-    void should_return_confirmation() {
+    @ParameterizedTest
+    @EnumSource(value = YesOrNo.class, names = { "NO", "YES" })
+    void should_return_confirmation(YesOrNo yesOrNo) {
```
neha-aggarw, line 42: can you please change the name of the parameter as 'isAda'?

neha-aggarw (changes requested):

src/functionalTest/resources/scenarios/RIA-412-RIA-1571-list-case-confirmation.json
```diff
@@ -7,15 +7,16 @@
       "eventId": "listCase",
       "state": "prepareForHearing",
       "caseData": {
-        "template": "minimal-appeal-submitted.json"
+        "template": "minimal-appeal-submitted.json",
+        "isAcceleratedDetainedAppeal": "No"
```
ThomasKKC, line 11: Can you add another json (copy of this) where isAcceleratedDetainedAppeal is "Yes"? and then return the correct expectation body

neha-aggarw: There is already one for ADA: minimal-ada-appeal-submitted.

neha-aggarw: Also looking at it, this test is not ada specific. So this should stay the same with isAcceleratedDetainedAppeal = No to the template for default normal flow.
but we should have another test for ADA flow.

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/postsubmit/ListCaseConfirmationTest.java
JamieHMCTS: I have changed the name of parameter yesOrNo to isAda

ThomasKKC (approved):

neha-aggarw (approved):
LGTM - the functional test will be added in a separate PR.

PR #1223, opened 2023-01-12
RIA-6508: Remove AA* fields from the asylumCaseDefinitions

neha-aggarw:
- Since the order of the screens (For Appellant details) is same, we will use the same fields for the non-AAA and AAA flows.
- Updated handlers, tests and functionalTestTemplate.

```
[ ] Yes
[x] No
```


ThomasKKC (approved):

JamieHMCTS (approved):
LGTM

PR #1244, opened 2023-01-24
RIA-6572 Wire AIP into the cancelling of automatic end of appeal

Marcellofabbri:
[RIA-6572](https://tools.hmcts.net/jira/browse/RIA-6572)

### Change description ###
- An AIP payment doesn't seem to use the `updatePaymentStatus` event to update the payment status, but it's resolved within the `paymentAppeal` event, which needs to be the qualifying event for the AIP journey for the CancelAutomaticEndAppealPaidConfirmation handler to update the trigger schedule to 100 years

Example below:
- 15:03:35: case 1674572435640710 is submitted, `endAppealAutomatically` event scheduled for 2 minutes later (15:05:35)
- 15:03:58: case 1674572435640710 gets paid, `endAppealAutomatically` event re-scheduled for ~100 years later (15:05:35)
- nothing gets logged on ia-case-api between 15:03 and 15:06, when the `endAppealAutomatically` event was first supposed to be triggered (but wasn't, because of it being successfully rescheduled for a century later)

image

```
[ ] Yes
[x] No
```


augustio (approved):
LGTM.

PR #1296, opened 2023-03-02
RIA-6856 makeAnApplicationTypes available for some internal ada cases

Marcellofabbri:
[RIA-6856](https://tools.hmcts.net/jira/browse/RIA-6856)

### Change description ###
- `Transfer` should not be available. It's only available for non ADA cases from the 'prepareForHearing' state onwards
- `Adjourn` and `Expedite` are available for internal ADA cases in the 'awaitingRespondentEvidence' state, and continue to be available for non-ADA cases after listing
- `Update hearing requirements` is available after 'submitHearingRequirements' is done

There's quite a bit of code repetition in this PR because I didn't feel like deviating too much from the grid that the `switch` block sets up, so it's easily understandable what each state is supposed to incorporate. I guess I could make it drier but it'd deform the straightforward approach of the `switch` block.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/MakeAnApplicationTypesProvider.java
```diff
@@ -89,6 +105,21 @@ public DynamicList getMakeAnApplicationTypes(Callback<AsylumCase> callback) {
             case REASONS_FOR_APPEAL_SUBMITTED:
             case RESPONDENT_REVIEW:
             case SUBMIT_HEARING_REQUIREMENTS:
+
+                if (currentState == AWAITING_RESPONDENT_EVIDENCE
```
Marcellofabbri, line 109: I know this looks a bit clunky, but I couldn't put this block alone in the AWAITING_RESPONDENT_EVIDENCE option because I need the rest of the code (which is communal with the other states from line 100 to 107) to still run. 

I tried to put this block (lines 109-115) alone in the AWAITING_RESPONDENT_EVICENCE case and omitted the `break` keyword so that it'd also run the rest (lines 123-154), but code style checks don't allow fallthrough and flag it up as a code smell.

augustio (approved):
LGTM.

PR #1334, opened 2023-03-29
RIA-6907 Remove detained status event handlers and tests

Ryan-Nesbitt:

https://tools.hmcts.net/jira/browse/RIA-6907

### Change description ###

- Handler for 'Remove detained status' event to write the field 'appellantInDetention' to 'No' and clear all detention-related fields from case data
- New event ID added
- Event authorisations added for Admin Officer and Legal Officer
- Confirmation screen for the 'Remove detained status' event
- Units tests and functional test

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/functionalTest/resources/scenarios/RIA-6907-remove-detained-status.json
+++ b/src/functionalTest/resources/scenarios/RIA-6907-remove-detained-status.json
@@ -1,0 +1,39 @@
+{
+  "description": "RIA-6907 Remove detained status",
+  "request": {
+    "uri": "/asylum/ccdAboutToSubmit",
+    "credentials": "AdminOfficer",
+    "input": {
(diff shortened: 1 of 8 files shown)


hussain-faruk (approved):
Code LGTM. Something thing you could add, for the sake of completeness, is 2 extra tests to assert what happens when
1) Case is non detained
2) Case is ADA

Having said that, I have still given an approval as I know from ccd defs side the event will never be accessible if its an ADA case or non detained, and also because the code itself LGTM.

hussain-faruk (approved):

Dbarnes-v1 (approved):
lgtm

hussain-faruk (approved):
LGTM

PR #1340, opened 2023-03-31
RIA-6991 Hide EditAppealTypePreparer class for AIP journey

Marcellofabbri:
[RIA-6991](https://tools.hmcts.net/jira/browse/RIA-6991) 🐛 
Causes of the bug:
- AIP initiates a case and then uses `EDIT_APPEAL` for filling in all the data.
- On the LR journey `appealTypeForDisplay` is expected to be populated during `EDIT_APPEAL` and the class would throw an exception if not populated (which is expected, but only on the LR journey)
- 
### Change description ###
- Disabled the class for the AIP journey. `appealType` is set directly from there.

```
[ ] Yes
[x] No
```


colmlynch60 (dismissed):
LGTM

colmlynch60 (approved):
LGTM

Dbarnes-v1 (approved):
LGTM

PR #1347, opened 2023-04-11
RIA-6112 RIA-6133 Enable/Disable ftpa notification with feature flag

Marcellofabbri:
[RIA-6112](https://tools.hmcts.net/jira/browse/RIA-6112)
[RIA-6133](https://tools.hmcts.net/jira/browse/RIA-6133)

### Change description ###
- Use feature toggle to disable notifications for `applyForFTPAAppellant` if feature flag false && journey is AIP

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandler.java
@@ -16,6 +16,7 @@ import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.PreSubmitCallb
 import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.field.PaymentStatus;
 import uk.gov.hmcts.reform.iacaseapi.domain.handlers.HandlerUtils;
 import uk.gov.hmcts.reform.iacaseapi.domain.handlers.PreSubmitCallbackHandler;
+import uk.gov.hmcts.reform.iacaseapi.domain.service.FeatureToggler;
 import uk.gov.hmcts.reform.iacaseapi.domain.service.NotificationSender;
 
 @Component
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandler.java
```diff
@@ -126,7 +130,8 @@ private List<Event> getEventsToHandle(Callback<AsylumCase> callback) {
         }
 
         // TODO Get rid of this when notification is activated for AIP
-        if (HandlerUtils.isAipJourney(callback.getCaseDetails().getCaseData())) {
+        if (HandlerUtils.isAipJourney(callback.getCaseDetails().getCaseData())
+            && !featureToggler.getValue("aip-ftpa-feature", false)) {
```
augustio, line 134: You might want to remove the comment here and maybe only disable the notification if feature flag returns true, that is the if condition here would only consider the feature flag and not the journey type, since the related ticket is about activating AIP notification.

augustio (commented):
I think addressing the comment I left would be nice, otherwise, looks good to me.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/SendNotificationHandler.java
Marcellofabbri: I think both conditions still apply because APPLY_FOR_FTPA_APPELLANT triggers notifications both in the LR and AIP journey.

If we remove APPLY_FOR_FTPA_APPELLANT from the qualifying events based only on the flag, then when the flag is off all notifications will be switched off (including the ones in the LR journey, currently being sent to AO, LR and HO). With both conditions we can remove/keep the event based on the flag but only scoped to the AIP journey, so the LR journey stays untouched.

augustio: Yes you are very correct. Thanks.

augustio (approved):
LGTM.

PR #1366, opened 2023-05-02
RIA-7040 Remove LR event added to handler

Ryan-Nesbitt:

https://tools.hmcts.net/jira/browse/RIA-7040

### Change description ###

- Added the event 'removeLegalRepresentative' to the handler so that isAdmin flag is set to true and they have access to events clearing LR details
- LR email address can no longer be cleared as it is needed for the ccdSubmitted callback for notifications and NOC
- This can be updated through event 'updateLegalRepDetails'

```
[ ] Yes
[X] No
```


colmlynch60 (approved):
LGTM

PR #1368, opened 2023-05-02
changes for endAppealAutomatically

kalachandrasekar1:
https://tools.hmcts.net/jira/browse/RIA-7055

### Change description ###

```
[ ] Yes
[ ] No
```


charts/ia-case-api/values.preview.template.yaml
```diff
@@ -64,7 +64,7 @@ java:
       password: hmcts
 
 ia-case-payments-api:
-  enabled: false #set to true if testing against a local branch
+  enabled: true #set to true if testing against a local branch
```
filippo-possenti-version1, line 67: If the comment is correct, I presume this needs reverting?

charts/ia-case-api/values.preview.template.yaml
```diff
@@ -81,7 +81,7 @@ ia-case-payments-api:
       CCD_URL: http://{{ .Release.Name }}-ccd-data-store-api
 
 ia-case-documents-api:
-  enabled: false #set to true if testing against a local branch
+  enabled: true #set to true if testing against a local branch
```
filippo-possenti-version1, line 84: If the comment is correct, I presume this needs reverting?

Otherwise get rid of the comment (or update it accordingly).

config/owasp/suppressions.xml
```diff
@@ -117,7 +117,10 @@
 	<cve>CVE-2023-28708</cve>
 	<cve>CVE-2023-20861</cve>
     </suppress>
-    <suppress until="2023-04-18">
+    <suppress>
```
filippo-possenti-version1, line 120: I will accept removing the date here if we're absolutely certain that these vulnerabilities don't affect us.

Otherwise leave the date and just push it forward so we don't forget about it. Then make it sure we have a ticket on the board to look into them.

charts/ia-case-api/values.preview.template.yaml
```diff
@@ -9,11 +9,11 @@ java:
     POSTGRES_PASSWORD: hmcts
     POSTGRES_CONNECTION_OPTIONS: ""
     #Un comment APIs if you need to test a local version
-    #IA_CASE_DOCUMENTS_API_URL: https://${SERVICE_NAME}-documents-api.preview.platform.hmcts.net
+    IA_CASE_DOCUMENTS_API_URL: https://${SERVICE_NAME}-documents-api.preview.platform.hmcts.net
```
filippo-possenti-version1, line 12: According to the comment, I presume this needs to be reverted.

charts/ia-case-api/values.preview.template.yaml
```diff
@@ -9,11 +9,11 @@ java:
...
     #IA_CASE_NOTIFICATIONS_API_URL: https://${SERVICE_NAME}-notifications-api.preview.platform.hmcts.net
     IA_TIMED_EVENT_SERVICE_URL: "http://ia-timed-event-service-aat.service.core-compute-aat.internal/"
     IA_TIMED_EVENT_SERVICE_ENABLED: "true"
-    #IA_CASE_PAYMENTS_API_URL: https://${SERVICE_NAME}-payments-api.preview.platform.hmcts.net
+    IA_CASE_PAYMENTS_API_URL: https://${SERVICE_NAME}-payments-api.preview.platform.hmcts.net
```
filippo-possenti-version1, line 16: According to the comment, I presume this needs to be reverted.

charts/ia-case-api/values.preview.template.yaml
kalachandrasekar1: Before merge into master, it will be reverted

charts/ia-case-api/values.preview.template.yaml
kalachandrasekar1: Before merge it to master, it will be reverted

charts/ia-case-api/values.preview.template.yaml
kalachandrasekar1: Before merge it to master, it will be reverted

config/owasp/suppressions.xml
kalachandrasekar1: Updated. Please review it

src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/EndAppealHandlerTest.java
```diff
@@ -157,18 +155,6 @@ void handling_should_throw_if_cannot_actually_handle() {
...
-        when(callback.getCaseDetails()).thenReturn(caseDetails);
-        when(caseDetails.getCaseData()).thenReturn(asylumCase);
-        when(asylumCase.read(PAYMENT_STATUS, PaymentStatus.class)).thenReturn(Optional.of(PaymentStatus.PAID));
-        when(callback.getEvent()).thenReturn(Event.END_APPEAL_AUTOMATICALLY);
-
-        assertThatThrownBy(() -> endAppealHandler.handle(PreSubmitCallbackStage.ABOUT_TO_SUBMIT, callback))
-            .hasMessage("Cannot auto end appeal as the payment is already made!")
-            .isExactlyInstanceOf(IllegalStateException.class);
-    }
-
```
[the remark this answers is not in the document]
kalachandrasekar1: Updated. Please review it

javimarinc96 (dismissed):
LGTM

filippo-possenti-version1 (dismissed):

filippo-possenti-version1:
For reference when the ticket will be looked into again: this PR resulted in a P2 ticket being opened due to cases being closed even though they were paid.
The problem this PR was meant to resolve was one of log cleansiness: the logs are polluted with the exception that was replaced here. The exceptions are hit as the corresponding scheduled task was not deleted. The appropriate solution to this problem will entail finding out why the scheduled tasks are not deleted after the appeal has been paid and ensure they are, instead of modifying this logic.
Do not touch the exception that was deleted in this PR: it will cause mayhem.

PR #1376, opened 2023-05-09
SNI-3541: code fix applied and unit tests passed

javimarinc96:

https://tools.hmcts.net/jira/browse/SNI-3541

### Change description ###

Introduced change when changing the date due to a direction to maintain the same original clarifying questions asked by the case officer when updating fields to preserve the same questions on the record asked previously to change the due date

```
[ ] Yes
[x] No
```


mcareyma (dismissed):
It looks OK to me.

kalachandrasekar1 (dismissed):

kalachandrasekar1 (dismissed):

PR #1377, opened 2023-05-10
RIA-7052 Skip BE handling of supplementary data if citizen in AIP journey

Marcellofabbri:
[RIA-7052](https://tools.hmcts.net/jira/browse/RIA-7052)
([RIA-6516](https://tools.hmcts.net/jira/browse/RIA-6516))

### Change description ###
- With supplementary data being set from the AIP frontend, the setting of supplementary data from the backend can be muted when the case is AIP and the user is `"citizen"` (avoiding a 400 exception)
- changed integration test to verify that idamApi is now called two times instead of one (since a new call is being made to verify the user's role when choosing to skip the setting of supplementary data)

```
[ ] Yes
[x] No
```


rajeshthuraiyur (approved):
LGTM

PR #1384, opened 2023-05-15
RIA-7052 Skip supplementary data setting form BE if startAppeal + citizen

Marcellofabbri:
[RIA-7052](https://tools.hmcts.net/jira/browse/RIA-7052)

### Change description ###
- Skipped supplementary data setting when `START_APPEAL` as a Citizen user, as in that scenario supplementary data are set from the FE, and in other events the data-setting will be skipped because data won't be null anymore

```
[ ] Yes
[x] No
```


rajeshthuraiyur (approved):
LGTM

PR #1436, opened 2023-06-13
RIA-7397 markAsReadyForUtTransfer -> reinstateAppeal -> endAppeal incorrect content

hussain-faruk:
* Cleared appealReadyForUtTransfer field so that UI content on ended state is accurate when appeal not ended through markAsReadyForUtTransfer event
* Added test case for clearing the field

### Change description ###
* Cleared appealReadyForUtTransfer field so that UI content on ended state is accurate when appeal not ended through markAsReadyForUtTransfer event
* Added test case for clearing the field

```
[ ] Yes
[X] No
```


neha-aggarw (approved):

colmlynch60 (approved):
LGTM

PR #1445, opened 2023-06-19
RIA-7118: Write previous detention location in case data when 'Update…

neha-aggarw:
… Detention Location' Event completes.
- And enable notification for the same event.

```
[ ] Yes
[x] No
```


colmlynch60:
temp close

PR #1449, opened 2023-06-20
Revert "changes for endAppealAutomatically"

filippo-possenti-version1:
Reverts hmcts/ia-case-api#1368


filippo-possenti-version1:
This seems to have resulted in cases being struck out even though they were paid. Reverting.
Will need to question what was tested, as this PR was in Ready to Deploy.

PR #1457, opened 2023-06-22
RIA-7239 Add OOC decision type check when handling ADA appeals

hussain-faruk:
https://tools.hmcts.net/jira/browse/RIA-7239

### Change description ###
When appeal is changed from in-country (ADA) to OOC, the in-country fields such as isADA can remain. An extra check has been inserted to ensure the appeal is not treated as ADA if OOC decision type is present. If it is present, we know the appeal was moved from in-country to OOC.

### Testing ###
Tested and was able to move from ADA to OOC and save/submit appeal successfully. 

```
[ ] Yes
[X] No
```


colmlynch60 (approved):
LGTM

neha-aggarw (approved):

PR #1482, opened 2023-07-07
RIA-7353 set hearing channel type when listing without requirements

DennisHoV1:
https://tools.hmcts.net/jira/browse/RIA-7353

### Change description ###
-Added the new page of hearing channel type when trigger "listCaseWithoutHearingRequirements" event
-Retrieve the hearing channel type from common reference data
-Suppressed the CVE problem

```
[ ] Yes
[X] No
```

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -29,5 +29,6 @@
         <cve>CVE-2023-34981</cve>
         <cve>CVE-2023-20883</cve>
         <cve>CVE-2023-35116</cve>
+        <cve>CVE-2023-1370</cve>
     </suppress>
 </suppressions>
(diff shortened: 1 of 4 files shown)


Marcellofabbri (changes requested):
~Waiting for confirmation on what the call to ref-data is actually going to return (cos it returns nothing at the moment, and this code is based on the premise it'll return a list of hearing channels that we're currently mocking)~
Request seems fine, the returning body depends on what the ref-data service returns in each environment (will need the mocked call in wiremock changed), but this code should be ok as it is

`http://rd-commondata-api-demo.service.core-compute-demo.internal/refdata/commondata/lov/categories/HearingChannel?serviceId=BFA1&isChildRequired=Y`

augustio (approved):
LGTM.

Marcellofabbri (approved):

PR #1515, opened 2023-07-25
RIA-7641 Rename getter-like field to something non-getter-like

Marcellofabbri:
[RIA-7641](https://tools.hmcts.net/jira/browse/RIA-7641)

### Change description ###
- Renamed `getWitnessFullName` to `buildWitnessFullName`. If the method looks like a getter (starts with _-get_) it'll build an object with such field even if not declared in the class, and then it'll fail validation when passed to CCD.

```
[ ] Yes
[x] No
```


rajeshthuraiyur (approved):
LGTM

PR #1538, opened 2023-08-07
RIA-7545 non payment end appeal

ikirsanov:
https://tools.hmcts.net/jira/browse/RIA-7545

### Change description ###
Added END_APPEAL_AUTOMATICALLY to internal journey

```
[ ] Yes
[x ] No
```


ThomasKKC (approved):

neha-aggarw (commented):
Worth checking if we can go without this change. Because the event is triggered from timed service, does it even come to ia-case-api?

neha-aggarw (approved):

PR #1577, opened 2023-08-22
RIA-7635 -Manage interpreter details

robin-ieong:
https://tools.hmcts.net/jira/browse/RIA-7635

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/UpdateInterpreterDetailsMidEventHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        if (!canHandle(callbackStage, callback)) {
+            throw new IllegalStateException("Cannot handle callback");
+        }
+
+        final AsylumCase asylumCase = callback.getCaseDetails().getCaseData();
+        PreSubmitCallbackResponse<AsylumCase> response = new PreSubmitCallbackResponse<>(asylumCase);
+        Optional<List<IdValue<InterpreterDetails>>> optionalInterpreterDetailsList = asylumCase.read(INTERPRETER_DETAILS);
+
+        if (optionalInterpreterDetailsList.isPresent()) {
+            validateInterpretersEmailAddresses(optionalInterpreterDetailsList, response);
```
Marcellofabbri, line 50: I think if you make the field of type "Email" rather than "Text" in the definitions, CCD will take care of it automatically without the need for the backend to do the validation

Marcellofabbri (changes requested):

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/postsubmit/UpdateInterpreterDetailsConfirmation.java
```diff
@@ -0,0 +1,49 @@
...
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.AsylumCase;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.Event;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.Callback;
+import uk.gov.hmcts.reform.iacaseapi.domain.entities.ccd.callback.PostSubmitCallbackResponse;
+import uk.gov.hmcts.reform.iacaseapi.domain.handlers.PostSubmitCallbackHandler;
+
+@Component
+public class UpdateInterpreterDetailsConfirmation implements PostSubmitCallbackHandler<AsylumCase> {
+
+    public boolean canHandle(
```
Itay-Kirsh, line 14: Is this formatting normal? Why not have the parameters on the same line as the method name? Same for the next ones.

Itay-Kirsh: Also for lots of the statements in the methods, you spread them over multiple lines but is there a reason to do that?

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/UpdateInterpreterDetailsMidEventHandler.java
robin-ieong: Ah I see, think I might have seen them before but just forgot about them over time. I have removed my validation and use the one provided by CCD instead, as there are no specific validation requirements in the story yet.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/postsubmit/UpdateInterpreterDetailsConfirmation.java
robin-ieong: Yes you're right that they don't look normal, I should have refactored them after copying from other classes, thanks.

Marcellofabbri (approved):

PR #1590, opened 2023-08-29
RIA-TASK - Updated Bundling dependencies for Master

colmlynch60:
### Change description ###

Updated Bundling dependencies for Preview

```
[ ] Yes
[x] No
```


filippo-possenti-version1 (approved):
Looks good to me.

I would just ask to add this:
https://github.com/hmcts/cnp-jenkins-library/blob/c7f6ae83b6c0f1841441ef4c7a6757ddf5389a1c/README.md?plain=1#L275

As a cautelative measure given that we're increasing the number of pods we run.

PR #1616, opened 2023-09-07
SNI-4584 Fix duplicate endAppealAutomatically case events

alivenichoppa:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/SNI-4584

### Change description ###
Do not trigger "endAppealAutomatically" case event if the appeal is already ended (when payment status is pending)

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1619, opened 2023-09-07
SNI-4584 fix duplicate end appeal events

alivenichoppa:
https://tools.hmcts.net/jira/browse/SNI-4584

### Change description ###
Do not trigger "endAppealAutomatically" case event if the appeal is already ended (when payment status is pending)

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -297,6 +297,12 @@ dependencyManagement {
             entry 'json-smart'
         }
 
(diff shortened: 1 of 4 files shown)


alivenichoppa:
Duplicate PR. see https://github.com/hmcts/ia-case-api/pull/1616

PR #1651, opened 2023-09-21
DIAC-14 nightly fix

nilay913:
https://tools.hmcts.net/jira/browse/DIAC-80

### Change description ###

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -344,7 +364,7 @@ dependencies {
 
     implementation group: 'uk.gov.hmcts.reform', name: 'logging', version: versions.reformLogging
     implementation group: 'uk.gov.hmcts.reform', name: 'logging-appinsights', version: versions.reformLogging
-    implementation group: 'uk.gov.hmcts.reform', name: 'properties-volume-spring-boot-starter', version: '0.0.4'
+    implementation group: 'uk.gov.hmcts.reform', name: 'properties-volume-spring-boot-starter', version: '0.1.0'
```
KleoG, line 367: might need to look into this one as i think its now deprecated

security.sh
```diff
@@ -1,13 +0,0 @@
-#!/usr/bin/env bash
```
KleoG, line 1: Is the security.sh needed for the jenkins security checks?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-all.zip
```
KleoG, line 3: Any changes between 7.4 and 7.6?

KleoG (approved):

security.sh
nilay913: when security.sh isn't found, it pulls it from the HMCTS repo directly so uses the most up to date one

gradle/wrapper/gradle-wrapper.properties
nilay913: just added support for java 18 and 19, and some other dependency format supports, nothings been removed so shouldn't affect our current code

build.gradle
nilay913: this would be a larger ticket as we may need to apply some config stuff for spring, also potential that its already been done and the above is no longer used, I just updated it to match other repos

KleoG (approved):

PR #1665, opened 2023-09-26
RIA-7657 Incorrectly removing all interpreter flags when no witness

cmccarthyHMCTS:
Incorrectly removed interpreter flags when no witnesses and this removed appellant flags
also on first event trigger for review update requirements only 1 witness would have its flags wrote as I was not rereading the flags


rajeshthuraiyur (approved):
LGTM

PR #1674, opened 2023-09-28
RIA-7703 create and remove interpreter case flags

cmccarthyHMCTS:
https://tools.hmcts.net/jira/browse/RIA-7703


rajeshthuraiyur (approved):
LGTM

PR #1721, opened 2023-10-13
RIA-7819 update isIntegrated when event triggered

cmccarthyHMCTS:
https://tools.hmcts.net/jira/browse/RIA-7819


rajeshthuraiyur (approved):
LGTM

PR #1734, opened 2023-10-23
RIA-7882 Trigger edit case listing event

MiguelMartinRojas:
### Jira link (if applicable)

RIA-7882 Trigger edit case listing event

### Change description ###

RIA-7882 Trigger edit case listing event


rajeshthuraiyur (approved):
LGTM

PR #1852, opened 2023-12-08
RIA-7781 - New Application type  Change hearing type

robin-ieong:
(https://tools.hmcts.net/jira/browse/RIA-7781)

### Checklist

- [ x] commit messages are meaningful and follow good commit message guidelines
- [x ] README and other documentation has been updated / added (if needed)
- [x ] tests have been updated / new tests has been added (if needed)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/MakeAnApplicationTypes.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/MakeAnApplicationTypes.java
@@ -17,8 +17,8 @@ public enum MakeAnApplicationTypes {
     UPDATE_HEARING_REQUIREMENTS("Update hearing requirements"),
     UPDATE_APPEAL_DETAILS("Update appeal details"),
     REINSTATE("Reinstate an ended appeal"),
-    OTHER("Other"),
-    CHANGE_HEARING_TYPE("Change hearing type");
+    CHANGE_HEARING_TYPE("Change hearing type"),
+    OTHER("Other");
 
     @JsonValue
     private final String value;
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/MakeAnApplicationTypes.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/MakeAnApplicationTypes.java
@@ -17,7 +17,8 @@ public enum MakeAnApplicationTypes {
     UPDATE_HEARING_REQUIREMENTS("Update hearing requirements"),
     UPDATE_APPEAL_DETAILS("Update appeal details"),
     REINSTATE("Reinstate an ended appeal"),
-    OTHER("Other");
+    OTHER("Other"),
+    CHANGE_HEARING_TYPE("Change hearing type");
 
     @JsonValue
     private final String value;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/MakeAnApplicationTypes.java
```diff
@@ -17,7 +17,8 @@ public enum MakeAnApplicationTypes {
     UPDATE_HEARING_REQUIREMENTS("Update hearing requirements"),
     UPDATE_APPEAL_DETAILS("Update appeal details"),
     REINSTATE("Reinstate an ended appeal"),
-    OTHER("Other");
+    OTHER("Other"),
```
rajeshthuraiyur, line 20: I would prefer to move this one above the OTHER type, as you know OTHER would come always at the end when they don't see anything that suits what they want.

Marcellofabbri (commented):
Unrelated to this PR, but it's worth checking if a PR needs to be opened for ia-case-notifications-api to add the new constant to the `MakeAnApplicationTypes` enum in that repository too. Some events that involve sending notifications may break if the application type is `changeHearingType` and the notification service doesn't include that value

robin-ieong:
> Unrelated to this PR, but it's worth checking if a PR needs to be opened for ia-case-notifications-api to add the new constant to the `MakeAnApplicationTypes` enum in that repository too. Some events that involve sending notifications may break if the application type is `changeHearingType` and the notification service doesn't include that value

Good point. I've debugged notification-api and can confirm it's using the BaseNotificationPersonalisation which doesn't require mapping the application type. There is no requirement to change the notification in this story but I've given a heads up to Simon to see if a customised notification is required like some of the other application types (if there is I expect it to be in a separate story). Thanks.

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/entities/MakeAnApplicationTypes.java
robin-ieong: updated, thanks

src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/MakeAnApplicationTypesProvider.java
```diff
@@ -126,6 +128,7 @@ public DynamicList getMakeAnApplicationTypes(Callback<AsylumCase> callback) {
                 values.add(new Value(LINK_OR_UNLINK.name(), LINK_OR_UNLINK.toString()));
                 values.add(new Value(JUDGE_REVIEW.name(), JUDGE_REVIEW.toString()));
                 values.add(new Value(OTHER.name(), OTHER.toString()));
+                values.add(new Value(CHANGE_HEARING_TYPE.name(), CHANGE_HEARING_TYPE.toString()));
```
Marcellofabbri, line 131: I just realized that `CHANGE_HEARING_TYPE` is missing from the `LISTING` block. It escaped me earlier when I checked the PR

Marcellofabbri (changes requested):

Marcellofabbri (approved):

rajeshthuraiyur (approved):
LGTM

PR #1870, opened 2023-12-18
RIA-8180 Use a correct case state after adjourn a case.

MiguelMartinRojas:
Jira link (if applicable)
https://tools.hmcts.net/jira/browse/RIA-8180

Change description
RIA-8180 Use a correct case state after adjourn a case.


augustio (approved):
LGTM.

PR #2537, opened 2025-04-29
DIAC-492 - sendDirectionWithQuestion - doesn't display correct error if direction due date in the past

guygrewal77:
### Jira link
https://tools.hmcts.net/jira/browse/DIAC-492


apereverzin1 (dismissed):

KleoG:
Hey @guygrewal77 i dont know if we want all of the code in the mid event handler, but only the date check and the rest of the code should stay in the other handler

alivenichoppa (approved):

PR #2559, opened 2025-05-12
DIAC-889 - Manage a fee option to be available for Admin/Caseworker

guygrewal77:
### Jira link
https://tools.hmcts.net/jira/browse/DIAC-889


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/ManageFeeUpdatePreparer.java
```diff
@@ -96,6 +95,7 @@ public PreSubmitCallbackResponse<AsylumCase> handle(
                 callbackResponse.addError("You cannot manage a fee update for this appeal");
                 break;
 
+            case PA:
```
nilay913, line 98: unsure if this change needs to be made as per the ticket - AC2 has been removed so I would expect it to still throw an error on a PA appeal if it hasn't been paid yet?

christianbola1 (approved):
Approve

PR #2634, opened 2025-07-10
DIAC-1237 ia-case-api bugs to resolve

david-gimelle:
### Jira link
https://tools.hmcts.net/jira/browse/DIAC-1352

### Change description
Fixing 2 potential race conditions. Case 1 and 2 of DIAC-1352

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [x] No

### Checklist

- [x] commit messages are meaningful and follow good commit message guidelines


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/OutOfTimeDecisionDetailsAppender.java
```diff
@@ -2,20 +2,24 @@
...
 
 @Service
-@Setter
-@Getter
 public class OutOfTimeDecisionDetailsAppender {
 
     private List<IdValue<OutOfTimeDecisionDetails>> allOutOfTimeDecisionDetails;
 
-    public List<IdValue<OutOfTimeDecisionDetails>> append(
+    public synchronized List<IdValue<OutOfTimeDecisionDetails>> getAllOutOfTimeDecisionDetails() {
```
KleoG, line 14: If we use allOutOfTimeDecisionDetails as a local variable in the append method. Do we need to use the getter and setter? im not sure if we need to hold the temporary state, but if we do we can use a request scope instead of synchronized?

KleoG (commented):
Do we need the getters and setter to make it a shared state? If we do, need to see if we use request scope or synchronised as synchronised limits access to the resource which could block some requests potentially

PR #2756, opened 2025-10-01
DIAC-1618 clearing payment status

tonymort:
https://tools.hmcts.net/jira/browse/DIAC-1618

Duplicate notifications sent to IRCs/Prisons after editing and submitting a remission appeal

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [ ] No

### Checklist

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/payment/FeesHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/payment/FeesHandler.java
@@ -107,12 +107,15 @@ public class FeesHandler implements PreSubmitCallbackHandler<AsylumCase> {
                 if (isRemissionsEnabled == YES && optRemissionType.isPresent()
                     && optRemissionType.get() == HO_WAIVER_REMISSION) {
                     setFeeRemissionTypeDetails(asylumCase);
+                    asylumCase.clear(PAYMENT_STATUS);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/EditAppealRemissionHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+        final AsylumCase asylumCase =
+            callback
+                .getCaseDetails()
+                .getCaseData();
+
+        Optional<RemissionType> remissionType = asylumCase.read(REMISSION_TYPE, RemissionType.class);
+
+        if (remissionType.isPresent()) {
+            asylumCase.clear(PAYMENT_STATUS);
```
sabahirfan, line 49: For late remmission scenario (after payment is done), we will have both payment status and remission type as valid scenario. please double check with business.

rajeshthuraiyur (changes requested):
I don' think we need to remove the payment status, we should the remission type to stop the non remission notification

PR #2958, opened 2026-01-12
DIAC-1475 call Home Office Integration API

ed-graham:


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/HomeOfficeReferenceHandler.java
```diff
@@ -59,6 +80,7 @@ public PreSubmitCallbackResponse<AsylumCase> handle(
 
         final AsylumCase asylumCase = callback.getCaseDetails().getCaseData();
 
+        // >>>ASK DAVID: why these restrictions on when the validation is run?<<< I THINK THEY MAY NEED TO BE REMOVED (at least some of them) because we always want to go through validation, don't we?
```
david-gimelle: I am not sure about this. Looks like a copy and past from somewhere else. But I wonder if it's possible to have a case without home office number? I don't know what is an internal case, but it could make sense it doesn't have an home office number. Same for someone out of the country, how they could have an home office reference?

david-gimelle (dismissed):

nilay913 (dismissed):

PR #3026, opened 2026-02-11
DIAC-2133 Fix notice of decision document not showing for rehydrated appeals

sgmselli:
### Jira link

https://tools.hmcts.net/jira/browse/DIAC-2133

### Technical issue

Technical Issue

- In the Start Appeal event, the field uploadTheNoticeOfDecisionDocs is mandatory for standard (paper) appeals.

- For rehydrated appeals, this field needs to be optional.

- In this PR: https://github.com/hmcts/ia-ccd-definitions/pull/2078/ a new field uploadRehydratedNod was introduced.

- uploadTheNoticeOfDecisionDocs → shown for non-rehydrated appeals (mandatory)

- uploadRehydratedNod → shown for rehydrated appeals (optional)

- The issue is that downstream APIs only read uploadTheNoticeOfDecisionDocs to retrieve the uploaded documents.

- For rehydrated appeals, the documents are stored in uploadRehydratedNod, so they are not picked up by the APIs.

### Solution Options

**1. Make the same field conditional in CCD (Preferred but not possible)**

- Make uploadTheNoticeOfDecisionDocs mandatory for non-rehydrated appeals and optional for rehydrated appeals.

- This solution is NOT possible because we cannot keep two of the same variable in ccd. We must create another to have one optional and one mandatory.

**2. Copy value in event handler (Preferred)**

- Create a new Presubmit handler for a rehydrated appeal that triggers on start or edit appeals that checks if uploadRehydratedNod is populated and uploadTheNoticeOfDecisionDocs is not. 

- If so, we populate uploadTheNoticeOfDecisionDocs with uploadRehydratedNod value and then clear uploadRehydratedNod value.

**3. Update consumers to read both fields (Least preferred)**

- Whenever we read uploadTheNoticeOfDecisionDocs, we can add fallback logic to look at uploadRehydratedNod if value empty

- This solution is less preferred because we would need more changes in the codebase, will be more difficult to maintain if we add more uploadTheNoticeOfDecisionDocs in code, and is more error prone.

### Solution

This PR implements solution 2.

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/RehydratedNoticeOfDecisionUploadHandlerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/iacaseapi/domain/handlers/presubmit/RehydratedNoticeOfDecisionUploadHandlerTest.java
@@ -2,24 +2,18 @@ package uk.gov.hmcts.reform.iacaseapi.domain.handlers.presubmit;
 
 import static org.assertj.core.api.Assertions.*;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.*;
 import static uk.gov.hmcts.reform.iacaseapi.domain.entities.AsylumCaseFieldDefinition.*;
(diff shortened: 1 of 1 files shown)


rajeshthuraiyur (approved):
LGTM

PR #3151, opened 2026-04-23
DIAC-1476 Call Home Office Validation API

ed-graham:


charts/ia-case-api/values.aat.template.yaml
```diff
@@ -1,4 +1,6 @@
 java:
-  # Don't modify below here
+  # Don't modify below here unless you know what you are doing
   image: ${IMAGE_NAME}
-  ingressHost: ${SERVICE_FQDN}
\ No newline at end of file
+  ingressHost: ${SERVICE_FQDN}
+  environment:
+    HOME_OFFICE_VALIDATION_ENABLED: "false"
```
velimd, line 6: i think this can be without quotations.

velimd: same about the ones below.

david-gimelle (dismissed):

velimd (dismissed):

charts/ia-case-api/values.aat.template.yaml
ed-graham: I think the quotes make it a string rather than a Boolean.  I was just following the precedent set elsewhere, which seems to be to use strings.

david-gimelle (dismissed):

charts/ia-case-api/values.aat.template.yaml
velimd: i thought it was meant to be a boolean. theres an example here without the quotes
https://github.com/hmcts/ia-case-api/pull/3151/changes#diff-e84898026e82e611331e359f1f23f9badf19bfb600d539666ec2f84cfaf3db05R39

PR #3168, opened 2026-05-11
DIAC-2480 EG Trigger build

ed-graham:


src/main/java/uk/gov/hmcts/reform/iacaseapi/domain/service/HomeOfficeReferenceService.java
```diff
@@ -26,14 +35,27 @@ public HomeOfficeReferenceService(HomeOfficeApi<AsylumCase> homeOfficeApi) {
...
+        if (!homeOfficeAppellantsSerialised.isEmpty()) {
+            log.info("Deserialising and returning previously retrieved Home Office appellant data for case with Home Office reference {}.", hoReference);
+            try {
+                homeOfficeAppellants = mapper.readValue(
+                                                            homeOfficeAppellantsSerialised,
+                                                            new TypeReference<List<IdValue<HomeOfficeAppellant>>>() {}
+                                                       );
+                return homeOfficeAppellants;
+            } catch (Exception ex) {
+                log.error("Could not deserialise list of Home Office appellants from serialised string {} for case with Home Office reference {}:\n\n{}",
```
david-gimelle, line 56: Same suggestion as in the API service for log of interaction with Home Office Api. We should be careful to log the full stack trace to not lost the "unexpected" root cause of the issue. Often getMessage, they are very little information. like "" or "null". But probably that should be done on home office api service, not here.

david-gimelle (approved):

velimd (approved):

PR #3213, opened 2026-06-04
DIAC-2168 - Remove appellant surname from INFO log

alext-justice:
Surnames were being logged at INFO level which gets shipped to centralised logging. This is a data protection concern as PII should not be present in application logs. The caseId and caseReferenceNumber remain in the log for debugging purposes.


nilay913 (approved):

rajeshthuraiyur (approved):
LGMT

